PR #10, opened 2024-01-19
Redmine update

jirihofman:
# What
- [x] correctly handle arrays. As strings in inspector, parse them in `receive()`
- [x] Add `outputType` option to `IssuesFind` by adding `x-connector-pagination` extension
- [x] Improve some labels and their examples
- [x] Refactor updated openapi spec to use refs.
- [x] Improve test flow

## Monday
Some unrelated Monday changes needed here in order to make the CI pass. These changes were merged before CI checks were introduced.
- [x] eslint fix. No connector updated needed there.
- [x] test fix - ignore `outputType` checks


src/appmixer/redmine/core/IssueCreate/component.json
```diff
@@ -55,34 +57,44 @@
...
+                        "description": "ID of the parent issue.",
                         "path": "issue.parent_issue_id"
                     },
                     "issue|custom_fields": {
                         "title": "CustomFields",
-                        "type": "array",
+                        "type": "string",
                         "path": "issue.custom_fields",
+                        "originalType": "array",
                         "items": {
```
harshaio, line 83: @jirihofman Since you changed the type from an array to a string, do we still need to keep these items?

jirihofman: We don't need `items` anymore but that would need to be handled in the generator. As of now it keeps many other properties there if they are in openAPI specs even though Appmixer doesn't use them.

Definitely a good idea to start removing them in the generator so that our `component.json` files are smaller.

vtalas (approved):

PR #14, opened 2024-01-23
Leadspicker: update descriptions

vtalas:


src/appmixer/leadspicker/artifacts/openapi.json
```diff
@@ -3,7 +3,7 @@
     "info": {
         "title": "NinjaAPI",
         "version": "1.0.0",
-        "description": "",
+        "description": "Seamlessly capture leads from LinkedIn, optimize your email strategy with tailored outreach solution, and harness the power of email rotation for a guaranteed high delivery rate.",
```
harshaio, line 6: @vtalas I see that you have already created a patched JSON file with an updated description. Why do we need to update it here?

vtalas: that's the updated, original file is the openapi.orig.json

harshaio (approved):

PR #16, opened 2024-01-29
ringring init

jirihofman:
Issue: https://github.com/clientIO/appmixer-components/issues/1661

### Done
- [x] Send SMS
- [x] correctly format and process `timeScheduled` and `timeValidity`

### Notes

1. There is no `auth.js` file. The `apiKey` needs to be a part of the body of the POST request to `/sms/v1/message`. Due to this, there is no authentication section for the component. Appmixer is not keeping track of any connected accounts.
3. I used demo account 
image


src/appmixer/ringring/core/MessageApiSmsRequest_Post/MessageApiSmsRequest_Post.js
```diff
@@ -0,0 +1,85 @@
...
+
+        // eslint-disable-next-line no-unused-vars
+        const input = context.messages.in.content;
+
+        let url = lib.getBaseUrl(context) + '/sms/v1/message';
+
+        const headers = {};
+
+        const inputMapping = {
+            'apiKey': input['apiKey'],
```
harshaio, line 24: I noticed that the `apikey` is present in this file. Could you please move it to the `auth.js` file?

jirihofman: Please see https://github.com/clientIO/appmixer-components/issues/1661#issuecomment-1914767748 where I mentioned this. The problem is that OpenAPI spec [doesn't support](https://swagger.io/specification/#security-scheme-object) `apiKey` in `body`.

harshaio: Before releasing the generator, we need to improve it or change the code.

jirihofman: Based on today's discussion we don't need to do any changes to the generator. When specified like this with `apiKey` in `body` we simply won't use Authentication section for the component.

vtalas (approved):

PR #21, opened 2024-02-16
Naxai Init

jirihofman:
Issue https://github.com/clientIO/appmixer-components/issues/1662

Done
- [x] `AddOrCreateContact`
- [x] `GetAttributes`
- [x] `GetContactIdentifier`
- [x] test flow

Manual changes after our openApi generator run
- [x] auth.js
- [x] using headers https://docs.naxai.com/reference/authentication#credentials-in-header
- [x] dynamic fields for `AddOrUpdateContact`
- [x] caching `GetAttributes` (used in `AddOrUpdateContact`)


src/appmixer/naxai/auth.js
```diff
@@ -0,0 +1,46 @@
...
+            },
+
+            clientSecret: {
+                type: 'password',
+                name: 'Client Secret',
+                tooltip: 'Provide your client secret.'
+            }
+        },
+
+        accountNameFromProfileInfo: 'clientId',
```
harshaio, line 23: @jirihofman Please add another field for the account name in the authentication screen keeping clientId as a fallback name.

jirihofman: What would be the benefit of it? clientId identifies exactly the api credentials that are used by the component.

harshaio: Users can add custom account names for easier association with client IDs which are hard to rember.

jirihofman: But it can also introduce more problems when user names several `clientId`s in the same way, eg. "My account". And we don't have any way to do some unique check here. I think it is fine as it is. It's the same as with any other `apiKey` authentication where user provides string like these.

harshaio (approved):

vtalas (approved):

PR #24, opened 2024-02-21
add ClickUp components.

harshaio:
- [x] FindTasks
- [x] CreateTask
- [x] UpdateTask
- [x] DeleteTask

- https://github.com/clientIO/appmixer-components/issues/1686


src/appmixer/clickup/test-flow.json
jirihofman, line 1: Format it, please.

jirihofman (changes requested):
Is there a reason why there is no `openapi.json` which is then used to generate the components?

harshaio:
> Is there a reason why there is no `openapi.json` which is then used to generate the components?

There are currently no available OpenAPI specifications for ClickUp.

PR #38, opened 2024-03-06
Fix Each component to handle transforms properly

jirihofman:
Fix for https://github.com/clientIO/appmixer-components/issues/1721

# Before
image

# After
image

### DIFF
--- a/src/appmixer/utils/controls/Each/component.json
+++ b/src/appmixer/utils/controls/Each/component.json
@@ -2,7 +2,7 @@
   "name": "appmixer.utils.controls.Each",
   "author": "David Durman <david@client.io>",
   "description": "Loops over the incoming list and outputs a new message for each item.",
-  "version": "1.4.4",
+  "version": "1.4.1",
   "private": false,
   "inPorts": [
     {
--- a/src/appmixer/utils/controls/Each/package.json
+++ b/src/appmixer/utils/controls/Each/package.json
@@ -1,0 +1,7 @@
+{
+	"name": "appmixer.utils.controls.Each",
+	"version": "1.0.1",
+	"dependencies": {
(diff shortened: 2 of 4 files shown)


src/appmixer/utils/controls/Each/component.json
```diff
@@ -2,7 +2,7 @@
   "name": "appmixer.utils.controls.Each",
   "author": "David Durman <david@client.io>",
   "description": "Loops over the incoming list and outputs a new message for each item.",
-  "version": "1.4.4",
+  "version": "1.4.1",
```
martin-krcmar, line 5: ? shouldn't be `1.4.5`?

src/appmixer/utils/controls/Each/package.json
```diff
@@ -0,0 +1,7 @@
+{
+	"name": "appmixer.utils.controls.Each",
```
martin-krcmar, line 2: This is not used in the fix.

martin-krcmar (changes requested):

src/appmixer/utils/controls/Each/package.json
jirihofman: Right, that was only needed when doing `appmixer publish` from `Each/` folder. Should it be removed, though? The dependencies for components in `utils:controls` differ a lot.

martin-krcmar: Hmm, I don't see `uuid` in the package.json in appmixer/utils and I don't see any package.json in the appmixer/utils/controls 
So how could this work before? So keep it.
Fix the indentation, should be 2 or 4 spaces, looks you have tabs there.

martin-krcmar (changes requested):

PR #43, opened 2024-03-13
Trello: Use `context.httpRequest` instead of `node-trello` for GET HTTP requests

jirihofman:
Issue: https://github.com/clientIO/appmixer-components/issues/1728

# Done
- [x] Use `context.httpRequest` instead of `node-trello` for all GET requests
- [x] Use `context.httpRequest` instead of `node-trello` for all other requests
- [x] remove `getTrelloAPI`
- [x] remove `node-trello` dependency
- [x] add json schema condition for `boardId` for other components using it in the inspector
- [x] fix `sendArrayOutput` for file with no records
- [x] add test-flow for triggers


src/appmixer/trello/trello-commons.js
```diff
@@ -32,7 +29,7 @@ module.exports = {
             await context.sendJson(records, outputPortName);
         } else if (outputType === 'file') {
             // Into CSV file.
-            const headers = Object.keys(records[0]);
+            const headers = Object.keys(records[0] || {});
```
jirihofman, line 32: This fixes `Error: Cannot convert undefined or null to object` when there are no `records`.

vtalas (commented):
@jirihofman please update versions in component.jsons and there is a change in the common.js file so the service version should be updated as well

vtalas (approved):

PR #61, opened 2024-03-27
util.controls.onError: use SearchAfter  - port from the https://github.com/clientIO/appmixer-components/pull/1690

harshaio:
- Handled pagination with `searchAfter` ( Appmixer 6.0 introduces a new query parameter `searchAfter`)
- Used locks to handle a large number of logs.
- https://github.com/clientIO/appmixer-components/issues/1466

original PR https://github.com/clientIO/appmixer-components/pull/1690

### DIFF
--- a/src/appmixer/utils/controls/OnError/OnError.js
+++ b/src/appmixer/utils/controls/OnError/OnError.js
@@ -1,59 +1,94 @@
 'use strict';
 
-async function getErrors(context, now, errors = [], from = 0, size = 100) {
+async function fetchErrorsPage(context, lock, limit, logId = null, lastGridTimestamp = null, size = 100, errors = []) {
 
-    const result = await context.callAppmixer({
-        endPoint: '/logs',
-        method: 'GET',
-        qs: {
-            from,
-            size,
-            sort: 'gridTimestamp:asc',
-            query: `@timestamp:[${context.state.since} TO ${now}] AND severity:error`,
-            flowId: context.flowId
-        }
-    });
+    const qs = {
+        size: Math.min(limit, size),
+        query: 'severity:error',
+        flowId: context.flowId
+    };
 
-    errors = errors.concat(result.hits);
-
-    if (result.hits.length === 0 || result.hits.length < size) {
-        return errors;
+    if (logId && lastGridTimestamp) {
+        qs.searchAfter = [logId, lastGridTimestamp];
     }
 
-    return getErrors(context, now, errors, result.nextFrom, size);
+    while (true) {
+        try {
+            const result = await context.callAppmixer({
+                endPoint: '/logs',
+                method: 'GET',
+                qs
+            });
+
+            if (!(result.hits && result.hits.length > 0 && limit > result.hits.length)) {
+                break;
+            }
+
+            lock.extend(parseInt(context.config.lockTTL, 10) || 1000 * 60 * 2);
+            const lastLog = result.hits[result.hits.length - 1];
+            errors.push(...result.hits);
+
+            if (errors.length >= limit) {
+                break;
+            }
+
+            qs.searchAfter = [lastLog['_id'], lastLog['gridTimestamp']];
+        } catch (error) {
+            console.error('Error fetching errors:', error);
+            await context.log({ level: 'error', error: error.message || error });
+            break;
+        }
+    }
+
+    return errors;
 }
 
-/**
- * Add Flow name and component labels to those errors.
- * @param {Context} context
- * @param {Array} errors
- */
+// Adds labels to errors for better identification
 function addLabels(context, errors) {
 
-    return errors.map(err => {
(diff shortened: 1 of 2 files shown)


vtalas (approved):
comments, approval in the https://github.com/clientIO/appmixer-components/pull/1690

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -1,59 +1,94 @@
...
 
-    if (result.hits.length === 0 || result.hits.length < size) {
-        return errors;
+            if (errors.length >= limit) {
+                break;
+            }
+
+            qs.searchAfter = [lastLog['_id'], lastLog['gridTimestamp']];
+        } catch (error) {
+            console.error('Error fetching errors:', error);
```
martin-krcmar, line 37: REmove console.error. That can just cause troubles. You can keep the context.log.
Axios, for example, has huge exceptions, if you put them to stderr/stdout then logstash splits them line by line and creates tons of logs in Elastic.

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -1,59 +1,94 @@
...
+        size: Math.min(limit, size),
+        query: 'severity:error',
+        flowId: context.flowId
+    };
+
+    if (logId && lastGridTimestamp) {
+        qs.searchAfter = [logId, lastGridTimestamp];
+    }
+
+    while (true) {
```
martin-krcmar, line 15: Do not use `while (true)`. We discussed that many times.

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -1,59 +1,94 @@
...
+        const lockName = `errorProcessingLock-${context.componentId}`; // Define a unique lock name
+        let lock;
+        try {
+            lock = await context.lock(lockName, { ttl: context.config.lockTTL || 1000 * 60 * 5, maxRetryCount: 0 });
+            const { lastLogId, gridTimestamp } = context.state;
+            // ensure that if there is a limit set by context.config.limit
+            const limit = context.properties.limit
+                ? Math.min(
+                    parseInt(context.properties.limit),
+                    parseInt(context.config.limit) || Infinity
```
martin-krcmar, line 69: So there is `Infinity` limit? if not in `context.config.imit`? That is not good. In a combination with that `while(true)` you can get to a situation where you load tons of errors. The engine can easily generate tens of thousands of errors and you don't want to get all of them.
I would set the `default` (if no `context.config.limit`) to something like 1000.

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -1,59 +1,94 @@
...
-        const errors = await getErrors(context, now);
-        await context.sendArray(addLabels(context, errors), 'out');
-        return context.saveState({ since: now });
+                // Update state with the _id of the last log in the page
+                const newLastLogId = result[result.length - 1]['_id'];
+                const newGridTimestamp = result[result.length - 1]['gridTimestamp'];
+                return context.saveState({ lastLogId: newLastLogId, gridTimestamp: newGridTimestamp });
+            }
+        } catch (error) {
+            await context.log({ level: 'error', message: error.message || error });
```
martin-krcmar, line 86: There is no need for this construct. If you let the error propagate, the system will log it.

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -1,59 +1,94 @@
...
 
 module.exports = {
 
-    async start(context) {
+    async tick(context) {
+
+        const lockName = `errorProcessingLock-${context.componentId}`; // Define a unique lock name
+        let lock;
+        try {
+            lock = await context.lock(lockName, { ttl: context.config.lockTTL || 1000 * 60 * 5, maxRetryCount: 0 });
```
martin-krcmar, line 63: I don't see the lock being extended. If getting all logs takes more then 5 minutes (which could happen with that `while(true)` and `Infinite` limit. Then that could take more than 5 minutes. And if a second process gets here, after the TTL expires, you will get duplicates on the output port.

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -1,59 +1,94 @@
...
-        method: 'GET',
-        qs: {
-            from,
-            size,
-            sort: 'gridTimestamp:asc',
-            query: `@timestamp:[${context.state.since} TO ${now}] AND severity:error`,
-            flowId: context.flowId
-        }
-    });
+async function fetchErrorsPage(context, lock, limit, logId = null, lastGridTimestamp = null, size = 100, errors = []) {
```
martin-krcmar, line 3: I think the `searchAfter` feature is only in Appmixer 6.0. But I cannot see a version dependency here (bundle.json).

PR #67, opened 2024-04-02
google.drive: add "supportsAllDrives" option

harshaio:
- Added `supportsAllDrives` option to support both 'My Drives' and 'Shared Drives'(By default, it supports all drives)

- https://github.com/clientIO/appmixer-components/issues/1694


jerbob92:
Hi! I don't really understand why this option would be given to the user. Afaik the option tells Google that the **application** supports all drives, that the connection has been implemented in a way to handle that. I don't think a user would be able to tell what this option is going to do.

PR #72, opened 2024-04-04
fix: csv.addRow 

harshaio:
- Updated the locking mechanism to avoid race conditions
- [x] add quotas
- [ ] create new component `AddRows` which is able to accept an array of rows, possibly mark the `AddRow` deprecated (?) (will be in another PR)


src/appmixer/utils/csv/AddRow/AddRow.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
             parseNumbers,
             parseBooleans
         });
-
+        const lock = await context.lock(fileId, { retryDelay: 1000 });
```
vtalas, line 26: There has to be the try-finally block as if any of the code below fails and throws an exception, the `lock` remains locked, right?

vtalas (changes requested):

src/appmixer/utils/csv/AddRow/AddRow.js
```diff
@@ -48,13 +49,18 @@ module.exports = {
                     rowAsArray[i] = '';
                 }
             }
-
+            lockExtendInterval = setInterval(async () => {
+                if (lock) {
+                    await lock.extend(parseInt(context.config.lockExtendTime, 10) || 1000 * 60 * 1);
+                }
+            }, 30000);
```
martin-krcmar, line 56: the `interval` should be either configurable or somehow related to the `context.config.lockExtendTime`. If you change the `lockExtendTime` to 10s, then the interval will not work correctly. Plus the `TTL` should be configurable, too.

src/appmixer/utils/csv/AddRow/AddRow.js
```diff
@@ -24,8 +24,9 @@ module.exports = {
             parseBooleans
         });
         let lock;
+        let lockExtendInterval;
         try {
-            lock = await context.lock(fileId, { retryDelay: 1000 });
+            lock = await context.lock(fileId, { ttl: 1000 * 60 * 1, maxRetryCount: 0 });
```
martin-krcmar, line 29: is there a reason for the `maxRetryCount: 0`? Because that is a good feature. Imagine you have 5 messages going into this component. The first one gets the lock and starts modifying the csv. With retryCount 0, all the others get rejected and will be retried in a minute. With a retryDelay and retryCount, the 4 messages must not be directly rejected/put-to-retry-mechanism, but can wait here for a bit. When the first one releases the lock, the second will get it with and so on. Processing all those 5 messages can be faster (may not go to the `appmixer retry` mechanism, at all).

martin-krcmar (changes requested):

src/appmixer/utils/csv/CSVProcessor.js
```diff
@@ -369,7 +369,6 @@ module.exports = class CSVProcessor {
      */
     async addRow(newRow, closure) {
 
-        const lock = await this.context.lock(this.fileId);
```
martin-krcmar, line 372: why was the lock removed from the CSVProcessor? I am not sure this is correct. The lock is now only in the AddRow. What if other component calls CSVProcessor.addRow? Looks like it is called only in AddRow. But it breaks the pattern, that all the CSVProcess methods do the locking inside. 

Another thing, isn't this issue in the other components, like AddColumn which uses EditCSV, where the lock isn't being extended, either ...

src/appmixer/utils/csv/AddRow/AddRow.js
```diff
@@ -23,34 +23,47 @@ module.exports = {
...
+                const item = rowAsArray[i];
+                if (item === undefined || item === null) {
+                    rowAsArray[i] = '';
+                }
             }
+            lockExtendInterval = setInterval(async () => {
+                if (lock) {
+                    await lock.extend(parseInt(context.config.lockExtendTime, 10) || 1000 * 60 * 1);
+                }
+            }, parseInt(context.config.lockExtendInterval, 10) || 30000);
```
martin-krcmar, line 59: The default values are fine, interval is half the TTL. But If I set `lockExtendInterval` in the backoffice, then the `interval` is the same as the TTL, which is not correct. Could lead to situation, when you open the lock for a while.

src/appmixer/utils/csv/AddRow/AddRow.js
```diff
@@ -23,34 +23,47 @@ module.exports = {
             parseNumbers,
             parseBooleans
         });
+        let lock;
+        let lockExtendInterval;
+        try {
+            lock = await context.lock(fileId, {
+                ttl: parseInt(context.config.lockTTL, 10) || 1000 * 60 * 1,
+                retryDelay: 1000
```
martin-krcmar, line 31: 1s delay, isn't that a lot? I would keep the default values and if anything, then make it configurable. No hardcoded numbers in the components!

martin-krcmar (changes requested):

src/appmixer/utils/csv/CSVProcessor.js
```diff
@@ -367,34 +367,78 @@ module.exports = class CSVProcessor {
...
+
+            stream.on('data', (rowData) => {
+                writeStream.write(rowData.join(this.delimiter) + '\n');
+                if (closure(idx, rowData, false)) {
+                    writeStream.write(rowAsArray.join(this.delimiter) + '\n');
+                }
+                idx++;
+
+                // Extend lock periodically
+                if (idx % 1000 === 0) {
```
martin-krcmar, line 414: Discover_-_Elastic
this code generated over 56K of logs.

martin-krcmar: Should extend based on a `time` rather the row count, that is too often.

martin-krcmar, line 414: Discover_-_Elastic
this generated over 56K of logs. You should extend in a time interval, rather that every 1000th row.

martin-krcmar (changes requested):

vtalas (approved):

PR #74, opened 2024-04-05
fix: google.drive: name contains special chars

harshaio:
- [ ] Inspect all other components that may have the same problem.

### DIFF
--- a/src/appmixer/google/drive/bundle.json
+++ b/src/appmixer/google/drive/bundle.json
@@ -46,8 +46,11 @@
         "1.4.5": [
             "MoveFileOrFolder: fixes - prevent 'Increasing the number of parents is not allowed', added DEBUG info message - logging file metadata and request params. Set the DEBUG configuration key in the Backoffice to true to see the messages."
         ],
-        "1.4.7": [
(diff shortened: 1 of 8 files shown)


src/appmixer/google/drive/bundle.json
```diff
@@ -46,7 +46,7 @@
         "1.4.5": [
             "MoveFileOrFolder: fixes - prevent 'Increasing the number of parents is not allowed', added DEBUG info message - logging file metadata and request params. Set the DEBUG configuration key in the Backoffice to true to see the messages."
         ],
-        "1.4.6": [
+        "1.4.7": [
```
vtalas, line 49: @harshaio  please add a new release note, the 1.4.6 is out already, nevertheless, this is a regular update, we should describe the change. Note that you can check which version is released in the master branch in the private repo: https://github.com/clientIO/appmixer-components/tree/master

vtalas (changes requested):
@harshaio  please bump the version the CreateFolder, FindFileOrFolder and UploadFile component.json as well

vtalas (approved):

PR #77, opened 2024-04-10
Plugin approach: kafka connector

harshaio:
- kafka.NewMessage:  Added plugins to maintain a persistent Kafka connection until server restart or flow stoppage.

### DIFF
--- a/src/appmixer/kafka/common.js
+++ b/src/appmixer/kafka/common.js
@@ -8,11 +8,19 @@ const kafka = () => {
 
     const init = (options = {}) => {
 
-        const { clientId, brokers, ssl, saslMechanism, saslUsername, saslPassword } = options;
+        const {
+            clientId,
+            brokers,
+            ssl,
+            saslMechanism,
+            saslUsername,
+            saslPassword,
+            connectionTimeout = 10000
+        } = options;
         const config = {
             clientId,
             brokers: brokers.split(',').map(broker => broker.trim()),
-            connectionTimeout: 10000,
+            connectionTimeout: options.config?.connectionTimeout || connectionTimeout,
             ssl: ssl ? ssl.toLowerCase() === 'true' : !!saslMechanism,
             sasl: saslMechanism
                 ? {
--- a/src/appmixer/kafka/common.js
+++ b/src/appmixer/kafka/common.js
@@ -12,6 +12,7 @@ const kafka = () => {
         const config = {
             clientId,
             brokers: brokers.split(',').map(broker => broker.trim()),
+            connectionTimeout: 10000,
             ssl: ssl ? ssl.toLowerCase() === 'true' : !!saslMechanism,
             sasl: saslMechanism
                 ? {
--- a/src/appmixer/kafka/common.js
+++ b/src/appmixer/kafka/common.js
@@ -54,9 +54,9 @@ const kafka = () => {
     const createProducer = ({ createPartitioner = undefined, retry = undefined } = {}) => {
 
         if (!kafkaClient) throw new Error('Kafka client not created');
-        if (!producer) {
-            producer = kafkaClient.producer({ createPartitioner, retry });
-        }
+        if (producer) throw new Error('Producer already created');
+        producer = kafkaClient.producer({ createPartitioner, retry });
+        return producer;
     };
 
     // Connect producer, producer must exists(use createProducer).
@@ -74,10 +74,10 @@ const kafka = () => {
     };
 
     // Send message, Producer must be connected(use connectProducer).
-    const send = async ({ topic, messages }) => {
+    const send = async (params) => {
 
         if (!producer) throw new Error('Producer not created');
-        await producer.send({ topic, messages });
+        await producer.send(params);
     };
 
     const createConsumer = ({ groupId }) => {
--- a/src/appmixer/kafka/config.js
+++ b/src/appmixer/kafka/config.js
@@ -3,13 +3,9 @@
 module.exports = context => {
 
     return {
-        reconnectJob: {
+        syncConnectionsJob: {
 
-            schedule: context.config.reconnectSchedule || '0 */1 * * * *'
-        },
-        disconnectJob: {
-
-            schedule: context.config.disconnectSchedule || '0 */1 * * * *'
+            schedule: context.config.syncConnectionsSchedule || '0 */1 * * * *'
         }
     };
 };
--- a/src/appmixer/kafka/config.js
+++ b/src/appmixer/kafka/config.js
@@ -1,0 +1,15 @@
+'use strict';
+
+module.exports = context => {
+
+    return {
+        reconnectJob: {
+
+            schedule: context.config.reconnectSchedule || '0 */1 * * * *'
+        },
+        disconnectJob: {
+
+            schedule: context.config.disconnectSchedule || '0 */1 * * * *'
+        }
+    };
+};
--- a/src/appmixer/kafka/connections.js
+++ b/src/appmixer/kafka/connections.js
@@ -42,7 +42,12 @@ const processMessageData = (message) => {
 const addConnection = async (context, component, mode) => {
 
     const { topics, flowId, componentId, fromBeginning, authDetails, groupId } = component;
+    authDetails.connectionTimeout = context.config.connectionTimeout;
 
+    const connectionId = `${flowId}:${componentId}`;
+    if (openConnections[connectionId]) return; // Connection already exists, do nothing
+
+    await context.service.stateSet(connectionId, { ...component, mode });
     try {
         if (mode === 'consumer') {
             await addConsumerConnection(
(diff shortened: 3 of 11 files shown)


src/appmixer/kafka/connections.js
```diff
@@ -0,0 +1,91 @@
...
+                    await context.triggerComponent(
+                        flowId,
+                        componentId,
+                        processMessageData(message),
+                        { enqueueOnly: true }
+                    );
+                } catch (err) {
+                    if (err.message === 'Flow stopped.' || err.message === 'Missing flow.') {
+                        await connection.disconnect();
+                        delete openConnections[connectionId];
```
DavidDurman, line 67: Just call `removeConnection({ flowId, componentId })`

src/appmixer/kafka/config.js
```diff
@@ -0,0 +1,15 @@
...
+
+    return {
+        reconnectJob: {
+
+            schedule: context.config.reconnectSchedule || '0 */1 * * * *'
+        },
+        disconnectJob: {
+
+            schedule: context.config.disconnectSchedule || '0 */1 * * * *'
+        }
```
DavidDurman, line 13: Any reason why we need two jobs? In my view, the job should be named "syncConnectionsJob" and should just either connect those that were not added (compared to the service state) or disconnect those that are not longer needed. This is to me one function.

DavidDurman (changes requested):

src/appmixer/kafka/config.js
harshaio: It is possible to have the flexibility to schedule two jobs with the desired time intervals.

src/appmixer/kafka/jobs.js
```diff
@@ -0,0 +1,32 @@
...
+
+        const registeredComponents = await context.service.loadState();
+
+        for (const component of registeredComponents) {
+            await context.log('info', 'Reconnecting the component.');
+            await connections.addConnection(context, component.value);
+        }
+    });
+
+    await context.scheduleJob('disconnect', config.disconnectJob.schedule, async () => {
```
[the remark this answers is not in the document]
martin-krcmar: But there is still the problem with lots of connections. Even with concurrency, it can run for let's say a minute. You `loadState` and get 1000 connections, going through them and opening them. In the middle, one (many) flow is stopped, state is changed, but the job does not know about it and will open the connection. Yes, it won't do anything and if a message arrives, it will be stopped. But it would be good, not to open it in the first place.

Solution. Before actually opening/closing a connection from the job, you have to check again, if it still is in the `state`.

martin-krcmar (changes requested):

src/appmixer/kafka/connections.js
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                    resolveOffset(message.offset);
+                    await heartbeat();
+                }
+            }
+        });
+    } catch (error) {
+        // Remove the connection from openConnections if an exception occurs
+        await removeConnection({ flowId, componentId });
+        throw error; // Re-throw the error to propagate it to the NewMessage component
```
martin-krcmar, line 79: What if Kafka server A is down. And there is a running flow with a Kafka connector in it. The scheduler gets the list of connections and tries to connect to them. But because Kafka server A is down, the connect() will always fail.
This exception will propagate to the cron job and the job will fail, because the exception will cancel the `Promise.map` and therefore one component will block other connections.

src/appmixer/kafka/jobs.js
```diff
@@ -0,0 +1,58 @@
...
+                    await connections.addConnection(context, component.value);
+                }
+            };
+
+            // Map over the registered components with concurrency limit
+            await Promise.all(registeredComponents.map(component => limit(() => connectComponent(component))));
+
+            // Disconnect components that are in the existing connections but not in the service state
+            await Promise.all(existingConnections.map(connectionId => limit(async () => {
+                if (!registeredComponentsKeys.has(connectionId)) {
```
martin-krcmar, line 43: at this point, the Set `registeredComponentsKeys` can be quite old.
Scenario:
- Flow A with NewMessage is started, and the connection is created.
- There is a lot of connection (let's assume 1k)
- Flow A is stopped.
- Server is restarted
- The job triggers, going through the list, and opening connections.
- Flow A is started, the connection is open directly (API)
- The job just finished opening connections and is going to check if there are connections that should be closed.
- But it loaded the list of connections that should be open in the beginning, it does not reflect the reality, that the flow A was started in the meantime.
- It will find in the `existingConnection` the connection for flow A, but because it has an old `state` it will disconnect it!!
- For some time, the flow is running, user expecting messages, but nothing is happening.
- When the job triggers again, it will open the connection.


Solutions:
A. call `context.service.stateGet(key)` before actually opening/closing every connection to get the latest state
B. at least reload the `state` after finishing opening the connections.

martin-krcmar (changes requested):

src/appmixer/kafka/platform/SendMessage/SendMessage.js
```diff
@@ -2,23 +2,33 @@
...
     }
-};
+}
 
 module.exports = {
 
+    async stop(context) {
+
+        return context.callAppmixer({
+            endPoint: `/plugins/appmixer/kafka/connect/producer/${context.flowId}/${context.componentId}`,
```
martin-krcmar, line 21: I cannot see this API in the routes.js

martin-krcmar (changes requested):

src/appmixer/kafka/common.js
```diff
@@ -12,6 +12,7 @@ const kafka = () => {
         const config = {
             clientId,
             brokers: brokers.split(',').map(broker => broker.trim()),
+            connectionTimeout: 10000,
```
martin-krcmar, line 15: can be configurable?

src/appmixer/kafka/routes.js
```diff
@@ -0,0 +1,40 @@
...
+                const { flowId, componentId } = req.payload;
+                const { mode } = req.params;
+                if (mode === 'consumer') {
+                    await connections.removeConnection({ flowId, componentId }, mode);
+                }
+                await connections.addConnection(context, req.payload, mode);
+                if (mode === 'producer') {
+                    await connections.sendMessage({ flowId, componentId, payload: req.payload });
+                }
+                await context.service.stateSet(`${flowId}:${componentId}`, { ...req.payload, mode });
```
martin-krcmar, line 21: This is redundant. You use this API to send every message. So with every message you set the `state` (to the same value). Plus you save the `message` to the state, which is not correct. If you send 10k messages, you will trigger 10k mongo updates, that do nothing.

src/appmixer/kafka/routes.js
```diff
@@ -0,0 +1,40 @@
...
+        path: '/connect/{mode}',
+        options: {
+            handler: async (req, h) => {
+
+                const { flowId, componentId } = req.payload;
+                const { mode } = req.params;
+                if (mode === 'consumer') {
+                    await connections.removeConnection({ flowId, componentId }, mode);
+                }
+                await connections.addConnection(context, req.payload, mode);
```
martin-krcmar, line 17: Doesn't it create a new producer connection with every message sent?
I don't see this condition 
Plugin_approach__kafka_connector_by_harshaio_·_Pull_Request__77_·_clientIO_appmixer-connectors

in the addProducerConnection function.

martin-krcmar (changes requested):

src/appmixer/kafka/connections.js
```diff
@@ -0,0 +1,180 @@
...
+
+const addConnection = async (context, component, mode) => {
+
+    const { topics, flowId, componentId, fromBeginning, authDetails, groupId } = component;
+    authDetails.connectionTimeout = context.config.connectionTimeout;
+
+    const connectionId = `${flowId}:${componentId}`;
+    if (openConnections[connectionId]) return; // Connection already exists, do nothing
+
+    await context.service.stateSet(connectionId, { ...component, mode });
```
martin-krcmar, line 50: Because there are no locks or anything, I think the following can happen:

- A -> Each -> kafka.SendMessage
- the Each will generate 200 messages, for example
- you have a cluster of 1 node, for simplicity
- roughly 100 messages for kafka.SendMessage will be consumed at the same time (no quota on that component)
- they all will call the plugin API 
- and they all can get to this same line `await context.service.stateSet`
- the result will be, that you will open 100 connections to Kafka. Not one.

martin-krcmar (changes requested):

src/appmixer/kafka/connections.js
```diff
@@ -42,7 +42,12 @@ const processMessageData = (message) => {
 const addConnection = async (context, component, mode) => {
 
     const { topics, flowId, componentId, fromBeginning, authDetails, groupId } = component;
+    authDetails.connectionTimeout = context.config.connectionTimeout;
 
+    const connectionId = `${flowId}:${componentId}`;
+    if (openConnections[connectionId]) return; // Connection already exists, do nothing
```
harshaio, line 48: @martin-krcmar If there is an open connection, then it simply returns.

src/appmixer/kafka/connections.js
harshaio: but @martin-krcmar If there is an open connection, then it simply returns

src/appmixer/kafka/connections.js
```diff
@@ -0,0 +1,204 @@
...
+
+const addConnection = async (context, component, mode) => {
+
+    const { topics, flowId, componentId, fromBeginning, authDetails, groupId } = component;
+    authDetails.connectionTimeout = context.config.connectionTimeout;
+
+    const connectionId = `${flowId}:${componentId}`;
+
+    // If another connection operation is in progress, wait for it to complete
+    if (inProgress) {
```
martin-krcmar, line 52: has to be `inProgress[connectionId]`. Otherwise, you would combine multiple connections (from multiple SendMessage components. But adding connections for different `connectionId` can happen simultaneously.

martin-krcmar (changes requested):

martin-krcmar (approved):

PR #79, opened 2024-04-17
Zendesk chat: authentication using global keys 

vtalas:
new config `globalKeys` 
when truthy, `pre` renders the input for the `subdomain` only, otherwise renders `clientId`, `clientSecret` and `subdomain`

### DIFF
--- a/src/appmixer/zendesktickets/auth.js
+++ b/src/appmixer/zendesktickets/auth.js
@@ -3,7 +3,7 @@
 const lib = require('./lib');
 
 const hasGlobalKeys = function(context) {
-    return context.config?.globalKeys === 'true' || false;
+    return context.config?.globalKeys === true || context.config?.globalKeys === 'true' || false;
 };
 
 module.exports = {
--- a/src/appmixer/zendesktickets/auth.js
+++ b/src/appmixer/zendesktickets/auth.js
@@ -2,6 +2,10 @@
 
 const lib = require('./lib');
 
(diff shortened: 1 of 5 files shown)


src/appmixer/zendesktickets/auth.js
```diff
@@ -2,6 +2,10 @@
 
 const lib = require('./lib');
 
+const hasGlobalKeys = function(context) {
+    return context.config?.globalKeys === true || context.config?.globalKeys === 'true' || false;
```
harshaio, line 6: Need to handle case conversion like: true, TRUE, and True?

src/appmixer/zendesktickets/lib.js
```diff
@@ -12,9 +12,8 @@ module.exports = {
 
     getBaseUrl: function(context) {
 
-        let url = 'https://{subdomain}.{domain}.com';
-        url = url.replaceAll('{domain}', context.auth?.domain || context.config?.domain || 'zendesk');
-        url = url.replaceAll('{subdomain}', context.auth?.subdomain || context.config?.subdomain || 'example');
+        let url = 'https://{subdomain}.zendesk.com';
+        url = url.replaceAll('{subdomain}', context.subdomain || 'example');
```
harshaio, line 16: need to check `context.subdomain` and `context.auth.subdomain`

harshaio (approved):

PR #83, opened 2024-04-23
Xero triggers

jirihofman:
Issue: https://github.com/clientIO/appmixer-components/issues/1763

# Done
- [x] plugin
- [x] lock
- [x] quotas
- [x] `NewCustomer`
- [x] `UpdatedCustomer`
- [x] `NewInvoice`
- [x] `UpdatedInvoice`

# Notes
- we must process webhook payload "as is" (without parsing it as JSON) because the signature is for content that is not exactly `JSON.stringify(payload)`, see https://devblog.xero.com/keeping-your-integration-in-sync-implementing-xero-webhooks-using-node-express-and-ngrok-6d2976baac6d
- Xero splits the payload events randomly (for example for a 100 contacts imported from CSV). Sometimes it is 100 messages at once, sometimes split into eg 18 and 82.
- using optimized data retrieval: https://devblog.xero.com/new-filtering-on-the-accounting-api-b8acee14ec25
    - webhook component gets list of `resourceId` for each event (eg Contact)
    - retrieve Contacts by batches of 40 (for 200 contacts it is 5 HTTP requests instead of 200)

# Docs
https://developer.xero.com/documentation/guides/webhooks/overview/

### DIFF
--- a/src/appmixer/xero/accounting/NewContact/NewContact.js
+++ b/src/appmixer/xero/accounting/NewContact/NewContact.js
@@ -1,0 +1,33 @@
+'use strict';
+
+const { webhookHandler } = require('../../commons');
+
+module.exports = {
+
+    start: async function(context) {
+
+        const { tenantId } = context.properties;
+        const { componentId, flowId } = context;
+        const webhook = 'CONTACT.CREATE:' + tenantId;
+        await context.log({ step: 'Registering webhook', webhook });
+        // Subscribe to a static webhook events received via ../../plugin.js.
+        return context.service.stateAddToSet(webhook, { flowId, componentId, webhook });
+    },
+
+    stop: async function(context) {
+
+        const { tenantId } = context.properties;
+        const { componentId, flowId } = context;
+        const webhook = 'CONTACT.CREATE:' + tenantId;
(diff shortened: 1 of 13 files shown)


src/appmixer/xero/accounting/NewContact/NewContact.js
```diff
@@ -0,0 +1,33 @@
...
+
+const { webhookHandler } = require('../../commons');
+
+module.exports = {
+
+    start: async function(context) {
+
+        const { tenantId } = context.properties;
+        const { componentId, flowId } = context;
+        const webhook = 'CONTACT.CREATE:' + tenantId;
```
harshaio, line 11: I noticed that the webhook ID is not unique each time it is generated.

jirihofman: That is on purpose. Single tenant can have multiple `NewContact` triggers across many flows. All of them are stored in
```js
{'CONTACT.CREATE:553c5adf-3f13-42ab-82bc-1a5135f9c172': [
    {flowId: 'flow1', componentId: 'foo'},
    {flowId: 'flow1', componentId: 'bar'},
    {flowId: 'flow2', componentId: 'baz'},
]
```

harshaio (approved):

PR #84, opened 2024-04-25
util.controls.onError: use SearchAfter

harshaio:
- Handled pagination with `searchAfter` ( Appmixer 6.0 introduces a new query parameter `searchAfter`)
- Used locks to handle a large number of logs.
- https://github.com/clientIO/appmixer-components/issues/1466

original PR https://github.com/clientIO/appmixer-components/pull/1690

### DIFF
--- a/src/appmixer/utils/controls/OnError/OnError.js
+++ b/src/appmixer/utils/controls/OnError/OnError.js
@@ -1,47 +1,47 @@
 'use strict';
 
-async function fetchErrorsPage(context, lock, limit, logId = null, lastGridTimestamp = null, size = 100, errors = []) {
+async function fetchErrors(context, lock, limit, logId = null, lastGridTimestamp = null, size = 100) {
 
-    const qs = {
-        size: Math.min(limit, size),
-        query: 'severity:error',
-        flowId: context.flowId
-    };
+    let fetchedLogs = [];
+    let remainingLimit = limit;
 
-    if (logId && lastGridTimestamp) {
-        qs.searchAfter = [logId, lastGridTimestamp];
-    }
+    while (remainingLimit > 0) {
 
-    while (true) {
-        try {
-            const result = await context.callAppmixer({
-                endPoint: '/logs',
-                method: 'GET',
-                qs
-            });
+        const pageSize = Math.min(remainingLimit, size);
+        const qs = {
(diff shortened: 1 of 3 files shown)


vtalas (commented):
looks good to me, just one question though> 
`sort: ['_id:asc', 'gridTimestamp:asc']` - this will return records from the oldest to newest, right? Wouldn't be better to have it from the newest (`desc`)?

vtalas:
@harshaio 
have you considered using generators? like 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield
```js
function* fetchErrors(index) {
  while (index < 2) {
    yield result.hits;
    index++;
  }
}

const iterator = fetchErrors(0);

console.log(iterator.next().value);
// Expected output: 0

console.log(iterator.next().value);
// Expected output: 1
```

I'd speed up sending data and maybe save some memory (in case there is more then 1 page of records)

vtalas (approved):

src/appmixer/utils/controls/OnError/OnError.js
```diff
@@ -62,28 +57,22 @@ module.exports = {
...
-                return context.saveState({ lastLogId: newLastLogId, gridTimestamp: newGridTimestamp });
+                ? Math.min(parseInt(context.properties.limit), parseInt(context.config.limit) || Infinity)
+                : parseInt(context.config.limit) || 1000;
+
+            for await (const batch of fetchErrorsGenerator(context, lock, limit, lastLogId, gridTimestamp)) {
+                if (batch.length > 0) {
+                    // Update state with the _id of the last log in the page
+                    const newLastLogId = batch[batch.length - 1]['_id'];
+                    const newGridTimestamp = batch[batch.length - 1]['gridTimestamp'];
+                    await context.saveState({ lastLogId: newLastLogId, gridTimestamp: newGridTimestamp });
```
martin-krcmar, line 69: Always send a message to the output port and then save the `state`. In case of a crash after this line. The `tick` will be generated again, but it will read the `state` and one batch could be lost. In case of a crash, we rather duplicate data on the output port, rather than skip something.

martin-krcmar (changes requested):

vtalas (approved):

PR #86, opened 2024-04-25
utils.csv.AddRows

harshaio:
- A new component `AddRows` to support batch insertion of rows


src/appmixer/utils/csv/AddRows/component.json
```diff
@@ -0,0 +1,96 @@
...
+                        "index": 2,
+                        "defaultValue": ",",
+                        "tooltip": "A character to use as a delimiter between columns."
+                    },
+                    "rows": {
+                        "type": "text",
+                        "label": "Rows",
+                        "index": 3,
+                        "defaultValue": "[]",
+                        "tooltip": "The rows to append."
```
vtalas, line 64: it's not clear what format should we use. It has to be a JSON array so it would good to mention it, here, maybe with a simple example

src/appmixer/utils/csv/AddRows/AddRows.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+const CSVProcessor = require('../CSVProcessor');
+
+module.exports = {
+
+    async receive(context) {
+
+        const { withHeaders } = context.properties;
```
vtalas, line 8: it's not used here (there is no definition in the compoentn.json either), I think we can remove `withHeaders` completely

vtalas (changes requested):
I noticed there are not much changes comparing the `addRows` and `addRows` in tge CSVProcessor.js. It should be fairly easy to refactor to the use just one method, `addRows` is the obvious choice for me. 

maybe the processor.addRow might be refactored as something like 
```
    async addRow({ row, rowWithColumns }, closure) {
        this.addRows([row], closure);
    }
```


Also, I think the `context` is undefined here: 
https://github.com/clientIO/appmixer-connectors/blob/dev/src/appmixer/utils/csv/CSVProcessor.js#L384

src/appmixer/utils/csv/AddRows/AddRows.js
vtalas: btw, why is `withHeaders` is not implemented here?

src/appmixer/utils/csv/bundle.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appmixer.utils.csv",
-  "version": "1.2.5",
+  "version": "1.2.6",
```
vtalas, line 3: should be 1.3.0 as we're adding new features

vtalas:
@harshaio  please add the     `"engine": ">=5.0",` as it uses `context.replaceFileStream` introduced in the version 5 (same as here https://github.com/clientIO/appmixer-connectors/blob/dev/src/appmixer/google/youtube/bundle.json#L4)

jirihofman (changes requested):
Add a `test-flow.json` please. Ideally for both `AddRow` and `AddRows`.

vtalas:
please add the quota for user:

```
        {
            limit: 5,
            throttling: 'limit-concurrency',
            queueing: 'fifo',
            resource: 'loadFiles',
            scope: 'userId'
        },
```

vtalas (approved):

PR #100, opened 2024-05-21
Zendesk tickets - fix global auth keys

vtalas:
- fix when the `globalKeys` is set to `false`

### DIFF
--- a/src/appmixer/zendesktickets/auth.js
+++ b/src/appmixer/zendesktickets/auth.js
@@ -3,7 +3,8 @@
 const lib = require('./lib');
 
 const hasGlobalKeys = function(context) {
-    return context.config?.globalKeys === true || context.config?.globalKeys?.toLowerCase() === 'true' || false;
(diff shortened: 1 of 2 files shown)


src/appmixer/zendesktickets/auth.js
```diff
@@ -3,7 +3,8 @@
 const lib = require('./lib');
 
 const hasGlobalKeys = function(context) {
-    return context.config?.globalKeys === true || context.config?.globalKeys?.toLowerCase() === 'true' || false;
+    const globalKeys = context.config?.globalKeys;
+    return globalKeys === true || typeof globalKeys === 'string' && globalKeys.toLowerCase() === 'true' || false;
```
martin-krcmar, line 7: @vtalas do you want to increase version in the bundle.json?

martin-krcmar (changes requested):

jirihofman (approved):

PR #102, opened 2024-05-29
OneDrive: Add redirectUri to Picker Component

danlang-impartner:
Fix: Added new redirectUri to OneDrivePicker controller so its passed into the new change in the picker input component


danlang-impartner:
Hey first time trying this out where we contribute, so let us know if this works!

jirihofman (approved):
Thanks @danlang-impartner for the contribution! It looks good, the only thing missing was a version bump for the whole bundle as well as the component. We understand that this is something not cover thoroughly in our docs, but we have plans to improve it.

PR #117, opened 2024-06-26
google.gmail: Add and Update activities in Gmail connector

sayam-nasir:
**NewEmail** Trigger
- Bring New Emails from Inbox if no Label is selected
- filter Emails based on selected filters

**NewAttachment** Trigger
- Bring New Attachments from Inbox if no Label is selected
- filter Emails based on selected filters and fetch their attachments

* **ListSignatures** Action
- [x] List default signature from Gmail

**SendEmail** Action
- [x] Added input support for CC
- [x]  Added input support for BCC
- [x]  Added input support for Labels. Labels can be selected from a dropdown and assigned to the email.
- [x]  Added input support for Signature. Signature can be selected from a dropdown and assigned to the email.

**CreateDraft** Action
- [x]  Added input support for CC
- [x]  Added input support for BCC
- [x]  Added input support for Labels. Labels can be selected from a dropdown and assigned to the email.
- [x]  Added input support for Attachments
- [x]  Added input support for Signature. Signature can be selected from a dropdown and assigned to the email.

* **ListAttachments** - (required for the Get attachment)_
- [x] * **RemoveLabelFromEmail** Action
- [x] * **ReplyEmail** Action
- [x] * **FindEmail** Action
- [x] * **MakeAPICall** component
- [x] move dependencies from the google/google-commons.js to google/gmail/gmail-commons.js
- [x] fix streams(GetAttachment)  _Gmail can have a total of 25mb in attachments. there is no way to get the attachment other than a base64 encoded string, Tested and works well without having to get a stream directly from the gmail API_ 
- [x] save to file (NewAttachment) _saveFileStream used in place of saveFile_
- [x] ListEmails - update output type _Added output type First Email_


src/appmixer/google/gmail/SendEmail/component.json
```diff
@@ -7,7 +7,8 @@
         "service": "appmixer:google",
         "scope": [
             "https://www.googleapis.com/auth/gmail.compose",
-            "https://www.googleapis.com/auth/gmail.send"
+            "https://www.googleapis.com/auth/gmail.send",
+            "https://www.googleapis.com/auth/gmail.modify"
```
sayam-nasir, line 11: @vtalas By default gmail.send does not support attaching labels to an email. The labels can only be attached after the email is sent by doing gmail.modify. And for that gmail.modify scope is needed. I have added this scope here. If the logic seems right, we can go ahead with this, otherwise we can remove the label attaching part from the activity

src/appmixer/google/gmail/CreateDraft/CreateDraft.js
```diff
@@ -22,15 +23,36 @@ module.exports = {
...
+            const fileInfo = await context.getFileInfo(fileId);
+            const fileStream = await context.getFileReadStream(fileId);
+            return { filename: fileInfo.filename, content: fileStream };
+        });
+
+        console.log('Mail object before building:', mail); // Debugging
+
         return new Promise((resolve, reject) => {
-            mailcomposer(mail).build((err, email) => {
+            mailcomposer({ ...mail, keepBcc: true }).build((err, email) => {  // Added keepBcc option
```
sayam-nasir, line 48: @vtalas somehow bcc property is not being retained in both cases SendEmail and CreateDraft, Here in this line I have added this property keepBcc after looking here [https://nodemailer.com/extras/mailcomposer/#bcc](https://nodemailer.com/extras/mailcomposer/#bcc) I will try to make it work

src/appmixer/google/gmail/gmail-commons.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const BASE_URL = 'https://gmail.googleapis.com/gmail/v1';
+
+module.exports = {
+    async fetchData(context, endpoint, options = {}) {
```
vtalas, line 6: what about

```js
       async fetchData(context, endpoint, {
        method = 'GET',
        params = {},
        headers = {}
    } = {}) {

        const options = {
            method,
            url: `${BASE_URL}${endpoint}`,
            headers: {
                Authorization: `Bearer ${context.auth.accessToken}`,
                ...headers
            },
            params
        };

        return await context.httpRequest(options);
    }

```
 - it's clear what is expected in the `options` object
 - saved some space, I found some variable redundant 
 - now you don't need to send any `options` from the `ListSignatures`

src/appmixer/google/gmail/ReplyToEmail/ReplyToEmail.js
```diff
@@ -0,0 +1,103 @@
...
+            } else {
+                mail.html = context.messages.in.content.signature;
+            }
+        }
+
+        if (context.messages.in.content.thread) {
+            mail.threadId = context.messages.in.content.thread;
+        }
+
+        return new Promise((resolve, reject) => {
```
vtalas, line 65: it seems the code is the same as in the CreateDraft, could it be moved to the gmail-commons?

src/appmixer/google/gmail/ReplyToEmail/ReplyToEmail.js
```diff
@@ -0,0 +1,103 @@
...
+         * @type MailComposer
+         */
+        let mail = Object.assign({}, context.messages.in.content);
+        if (mail.from && mail.sender) {
+            mail.from = `${mail.sender} <${mail.from}>`;
+        } else if (mail.sender) {
+            mail.from = `${mail.sender} <${context.profileInfo.email}>`;
+        }
+
+        if (context.messages.in.content.cc) {
```
vtalas, line 30: this is redundant, right? 
as Object.assign put the `cc` and `bcc` into the object

also try to simplify, what about
(NOT TESTED)
```js 
 async receive(context) {

        /**
         * @type MailComposer
         */

        const {
            from = context.profileInfo.email,
            sender,
            cc,
            bcc,
            attachments = {},
            thread: threadId,
            labels,
            signature
        } = context.messages.in.content;

        const mail = {
            from: `${sender} <${from}>`,
            cc,
            bcc,
            attachments: await getAttachmentsContent(context, attachments),
            threadId
        };


        // Add signature to the email content if provided
        // TODO v.talas: add more comprehensive comment
        if (signature) {
            if (mail.html) {
                mail.html += `<br><br>${signature}`;
            } else if (mail.text) {
                mail.html = `${mail.text.replace(/\n/g, '<br>')}<br><br>${signature}`;
                delete mail.text;
            } else {
                mail.html = context.messages.in.content.signature;
            }
        }

        return new Promise((resolve, reject) => {
            const composer = mailcomposer(mail);
            composer.keepBcc = true;
            composer.build((err, email) => {
                if (err) {
                    return reject(err);
                }
                resolve(email);
            });
        }).then(async email => {
            // TODO v.talas - weird combination of promises and awaits 
            const result = await send({
                auth: commons.getOauth2Client(context.auth),
                userId: 'me',
                quotaUser: context.auth.userId,
                // URI-safe base64
                resource: {
                    raw: email.toString('base64').replace(/\+/gi, '-').replace(/\//gi, '_'),
                    threadId: mail.threadId
                }
            });

            if (labels && result.id) {
                await modify({
                    auth: commons.getOauth2Client(context.auth),
                    userId: 'me',
                    id: result.id,
                    resource: {
                        addLabelIds: context.messages.in.content.labels.AND.map(label => label.name)
                    }
                });
            } else {
                // TODO v.talas context.SendError is deprecated, throw an error instead 
                await context.sendError('Invalid email label, ' + JSON.stringify(email));
            }

            return context.sendJson(result, 'email');
        });
    }

```

src/appmixer/google/gmail/ListThreads/ListThreads.js
```diff
@@ -0,0 +1,31 @@
...
+const { promisify } = require('util');
+
+// GoogleApi initialization & promisify of some api function for convenience
+const gmail = GoogleApi.gmail('v1');
+
+module.exports = {
+    async receive(context) {
+
+        const listThread = promisify(gmail.users.threads.list.bind(gmail.users.threads));
+        const { threads } = await listThread({
```
vtalas, line 13: why the google library? I think we'd like to use the "fetchData" in the new files, the same as in the ListSignatures.js

vtalas (changes requested):
Email inputs: 
whenever there is an email on the intput, please use the label `Email`  (not the email id). In the tooltip, there should be a meaningful information. for example, AddLabelsToEmail  (https://github.com/clientIO/appmixer-connectors/pull/117/files#diff-ad818088569b85c4c21c6b53987e9b26ab64c13699a7ffadd08b093ff64af7cdR40) should have  "Select an email to be labeled" instead.  

Also, it should be possible to select the email - the same approach as in the ReplyToEmail: https://github.com/clientIO/appmixer-connectors/pull/117/files#diff-bf0ee1f3f67ec424bc996db0126a4a495f3536345c6be358d2933495b92f9b14 

Please review all components where there is email on the input - label, tooltip and dynamic input

src/appmixer/google/gmail/ListSignatures/component.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "appmixer.google.gmail.ListSignatures",
+    "author": "Sayam Nasir <sayam@client.io>",
+    "description": "Retrieves default signature from your Gmail.",
+    "private": true,
+    "version": "1.0.1",
+    "auth": {
+        "service": "appmixer:google",
+        "scope": ["https://www.googleapis.com/auth/gmail.settings.basic"]
```
vtalas, line 9: @sayam-nasir `gmail.settings.basic` is this scope needed? as it's a new scope and we don't have this approved, I'd require a re-validation of the scopes from Google. Could it be replaced by some existing scope?

vtalas (changes requested):
@sayam-nasir  could you please add the `firts` option, same as [here](https://github.com/clientIO/appmixer-connectors/pull/147/files#diff-b31e2b31db2d07b14556c4277d9787cf5bd6733139dd241d6df9b4fa64ddbad7R98)

it's the new output type for the array-like outputs

vtalas (changes requested):
for the output type: one at a time (e.g here https://github.com/clientIO/appmixer-connectors/pull/117/files#diff-c63bf6c649e79b5e32807df19e5f5f4201b3c8435a948e7b927f0219723d552dR62), use the context.sendArray, instead of the `for` cycle.  I know the `for`  cycle has been there already, but we've got an opportunity to refactor it now. 

also there is going to be a better readability where there is going to be explicitly set 
```
if (outputType === 'email') {...
if (outputType === 'emails') {
if (outputType === 'file') {
```
in the current version the `file` is mixed up with the `email`

src/appmixer/google/gmail/GetAttachment/GetAttachment.js
```diff
@@ -1,33 +1,19 @@
...
-            id: attachmentId,
-            messageId
+        const endpoint = `/users/me/messages/${messageId}/attachments/${attachmentId}`;
+        const attachment = await emailCommons.callEndpoint(context, endpoint, {
+            method: 'GET',
+            headers: { 'Content-Type': 'application/json' }
         });
 
-        const buffer = Buffer.from(attachment.data, 'base64');
+        const buffer = Buffer.from(attachment.data.data, 'base64');
```
vtalas, line 14: it's been here already, however we need to fix this, if possible. We should properly utilize streams working with files. The current versions loads the whole file into memory, but we should stream the file directly from the Google endpoint to our DB. Please investigate this

src/appmixer/google/gmail/ListAttachments/ListAttachments.js
```diff
@@ -0,0 +1,32 @@
...
+        const { messageId } = context.messages.in.content;
+        const endpoint = `/users/me/messages/${messageId}`;
+        const messageDetails = await emailCommons.callEndpoint(context, endpoint, {
+            method: 'GET',
+            params: {
+                format: 'full'
+            }
+        });
+        const normalizedEmail = emailCommons.normalizeEmail(messageDetails.data);
+        context.log({ normalizedEmail });
```
vtalas, line 15: please use format like `{ stage: "<name of the stage, e.g nomalizing emails> <log message>" }` or remove the context.log

src/appmixer/google/gmail/NewAttachment/NewAttachment.js
```diff
@@ -1,96 +1,55 @@
...
+            if (isLabelsEmpty || labels.some(label => email.labelIds.includes(label.name))) {
+                return downloadAttachments(auth, context, email);
+            }
+            return [];
         });
 
         attachments = attachments.reduce((a, b) => a.concat(b), []);
 
         let saved = await Promise.map(attachments, attachment => {
             return context.saveFile(
```
vtalas, line 50: context.saveFile  is deprecated, use the `context.saveFilestream` instead. 
also please use the streams as described in the comment for the `GetAttachment`

vtalas (commented):
remove the appmixer.google.zip from the PR

vtalas (changes requested):
# scopes:

**ReplyToEmail**
https://www.googleapis.com/auth/gmail.modify

**SendEmail**
https://www.googleapis.com/auth/gmail.modify

**CreateDraft**
https://www.googleapis.com/auth/gmail.modify


# author:
author in all component.jsons: "Appmixer <info@appmixer.com>"

src/appmixer/google/gmail/ListSignatures/component.json
sayam-nasir: replaced it with the scope <https://www.googleapis.com/auth/gmail.readonly>

src/appmixer/google/gmail/GetAttachment/GetAttachment.js
sayam-nasir: The only way files are imported from google endpoint is in the form of a base64 encoded string, the sum of total files/attachments size on an email can be no more than 25mb.

src/appmixer/google/gmail/NewAttachment/NewAttachment.js
sayam-nasir: replaced with saveFilestream

vtalas:
- [ ] - ListEmails + FindEmail => **FindEmails**
   - instead of the listEmails and FindEmail there is only one action **FindEmails**
   - query, limit -> array<Email ID> 
   - (??) labels, deleted/spam  include these inputs? 
 - [ ] - copy the auth.js google/auth.js -> google/gmail/auth.js
 - [ ] - remove Gmail related methods from the google/commons.js

vtalas (changes requested):
- SendEmail 
  - output labels: first char upper case, id -> Email ID 
image

- Find emails: add tooltip: "Note that each message resource contains only an id and a threadId. Additional message details can be fetched using the GetEmail component."

- labeling: GetEmail, NewEamil (maybe more?) output label: id -> Email ID. For example FindEmail labels the email id as "Email ID" 
   - same id, 3 types of labels 
    image 

  - similar to other ids and labels - for example NewLabel - there is output `id`. Please keep the logic of labeling ids - id -> Label ID

  - NewAttachment - fileId -> File ID
  - please revisit all labels and it's format

vtalas (approved):

PR #118, opened 2024-06-27
Kafka: NewMessage, SendMessage

DavidDurman:


src/appmixer/kafkadavid/platform/SendMessage/SendMessage.js
```diff
@@ -0,0 +1,65 @@
...
+                    partition,
+                    timestamp: timestamp && new Date(timestamp).getTime(),
+                    headers: headers && JSON.parse(headers)
+                }
+            ],
+            acks,
+            timeout
+        };
+
+        await context.callAppmixer({
```
martin-krcmar, line 56: Scenario:
- a new pod is added, or engine is restarted
- message from Rabbit is consumed and gets here
- the job in Kafka plugin did not run yet, it's scheduled to run every minute - therefore the connection is not created
- POST to /send 
- the connections.sendMessage is triggered and there is a condition ` if (!producer) return; // Connection doesn't exist, do nothing` and your message is lost forever.

src/appmixer/kafkadavid/connections.js
```diff
@@ -0,0 +1,177 @@
...
+
+    await producer.connect();
+    OPEN_CONNECTIONS[connectionId] = producer;
+};
+
+const sendMessage = async (flowId, componentId, payload) => {
+
+    const connectionId = `producer:${flowId}:${componentId}`;
+    const producer = OPEN_CONNECTIONS[connectionId];
+    if (!producer) return; // Connection doesn't exist, do nothing
```
martin-krcmar, line 140: It should throw an error, that errors get to the component and the message will be retried. Otherwise it is lost.

src/appmixer/kafkadavid/connections.js
```diff
@@ -0,0 +1,177 @@
...
+
+const addProducer = async (context, flowId, componentId, auth) => {
+
+    const connectionId = `producer:${flowId}:${componentId}`;
+    await context.service.stateSet(connectionId, {
+        flowId, componentId, auth
+    });
+    auth.connectionTimeout = context.config.connectionTimeout;
+    const client = initClient(auth);
+    const producer = client.producer();
```
martin-krcmar, line 130: What will happen if the producer crashes? The job will recreate it? I can see 
`consumer.on(consumer.events.CRASH...` but not the same for the producer.

src/appmixer/kafkadavid/platform/ListOpenConnections/ListOpenConnections.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const connections = require('../../connections');
```
martin-krcmar, line 3: There is a security issue, you went around the node cache by storing the connections to the `process`, but by calling this component, every user will be able to get all Kafka connections from all users in Appmixer.

src/appmixer/kafkadavid/jobs.js
```diff
@@ -0,0 +1,97 @@
...
+                    if (connectionId.startsWith('consumer')) {
+                        await connections.removeConsumer(context, conn.flowId, conn.componentId);
+                    } else {
+                        await connections.removeProducer(context, conn.flowId, conn.componentId);
+                    }
+                }
+            }));
+
+        } catch (error) {
+            await context.log('error', `Error occurred during connection sync job: ${error.message}`);
```
martin-krcmar, line 91: appmixer-1-gmlmw_·_Pod_·_Logs_·_OKD

I saw this error in the logs of `learning-gull-30931` tenant.

src/appmixer/kafkadavid/platform/ListOpenConnections/ListOpenConnections.js
DavidDurman: @martin-krcmar how would an end-user get this? Do you mean by using the "JavaScript Function" modifier? How can an end-user get access to the `process`?

martin-krcmar: I mean by adding this component ListOpenConnections to their flow. OnStart -> LIstOpenConnections -> SendEmail and you get to your mailbox all open connections.

DavidDurman: @martin-krcmar right, the `ListOpenConnections` is not going out, it was just a handy debugging tool. It's removed now.

vtalas (approved):

PR #120, opened 2024-07-03
Mongo: Remove tmp files when connection fails

jirihofman:
Fix for https://github.com/clientIO/appmixer-components/issues/1768

### Notes
- The issue exists when the connection is **not** successful.
- I tried using `tmp.setGracefulCleanup();` but it inly removes the files on process end. Not immediately.

### DIFF
--- a/src/appmixer/mongodb/common.js
+++ b/src/appmixer/mongodb/common.js
@@ -26,7 +26,23 @@ module.exports = {
             options.tlsAllowInvalidCertificates = true;
         }
         const client = new MongoClient(context.connectionUri, context.tlsCAFileContent && options);
-        await client.connect();
+        try {
+            await client.connect();
+        } catch (error) {
+            if (context.tlsCAFileContent) {
+                // Removing the temporary file and directory if the connection fails
+                await new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


src/appmixer/mongodb/common.js
```diff
@@ -26,7 +26,23 @@ module.exports = {
             options.tlsAllowInvalidCertificates = true;
         }
         const client = new MongoClient(context.connectionUri, context.tlsCAFileContent && options);
-        await client.connect();
+        try {
+            await client.connect();
+        } catch (error) {
+            if (context.tlsCAFileContent) {
+                // Removing the temporary file and directory if the connection fails
+                await new Promise((resolve, reject) => {
```
vtalas, line 34: perhaps it could be simplified as: 

` await fs.rm(path, { recursive: true, force: true });` 

using the `const fs = require('fs').promises;`

jirihofman: I wasn't able to make it work with `fs.promises` - the folder always remained even though the file was removed.

Instead I refactored the code to extract the function.

vtalas (approved):

PR #124, opened 2024-07-08
Slack: add UploadFile

jirihofman:
Issue https://github.com/clientIO/appmixer-components/issues/1580

Using https://slack.dev/node-slack-sdk/web-api

- [x] UploadFile
- [x] DeleteFile
- [x] GetFile

### Notes
- For `GetFile` we return only `file`. `comments` are [deprecated](https://api.slack.com/changelog/2018-05-file-threads-soon-tread) and not needed. Because of this there is also no need for inputs: `count, cursor, limit, page` as they are related to comments.

### DIFF
--- a/src/appmixer/slack/bundle.json
+++ b/src/appmixer/slack/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.slack",
-    "version": "2.2.0",
+    "version": "2.2.1",
     "engine": ">=6.0.0",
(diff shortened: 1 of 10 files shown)


jirihofman:
Added fix for tooltips from https://github.com/clientIO/appmixer-components/issues/1809#issuecomment-2216873221

vtalas (approved):

PR #135, opened 2024-07-22
Asana: fix assignee dynamic input

vtalas:
- follow-up for the https://github.com/clientIO/appmixer-connectors/pull/132 (added changelog)

### DIFF
--- a/src/appmixer/asana/tasks/ListUsers/ListUsers.js
+++ b/src/appmixer/asana/tasks/ListUsers/ListUsers.js
@@ -7,7 +7,7 @@ const commons = require('../../asana-commons');
  */
 module.exports = {
 
-    receive(context) {
+    async receive(context) {
 
         const generateOutputPortOptions = context.properties.generateOutputPortOptions;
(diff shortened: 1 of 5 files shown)


src/appmixer/asana/tasks/ListUsers/ListUsers.js
```diff
@@ -18,15 +18,23 @@ module.exports = {
...
+        let users;
+
+        try {
+            const res = await client.users.findByWorkspace(workspace);
+            users = res.data;
+        } catch (err) {
+            if (!context.properties.ignoreErrors) {
+                throw err;
+            }
+        }
```
jirihofman, line 30: What kind of errors does it throw?

I would like to see some kind of error logging here even when we are ignoring the errors.

vtalas: I think this is not necessary as this is for the UI only, I think you really don't case about the errors, If you do care, you can use the ListUsers without the ignoreErrors flag

jirihofman (approved):

PR #136, opened 2024-07-23
utils.files: add the ReadJson

vtalas:
reads the JSON file as a stream, emits records one by one, defined by the JSONPath: 

read json with array  JSON Path : `*`
image

read json - JSON path `address`

image


jirihofman (changes requested):
Bump `bundle.json` too

PR #137, opened 2024-07-30
Add OnAnyFlowComponentError

jirihofman:
Issue: https://github.com/clientIO/appmixer-components/issues/1561

- [x] add webhook URL to `WEBHOOK_FLOW_COMPONENT_ERROR` on `start`
- [x] remove webhook URL from `WEBHOOK_FLOW_COMPONENT_ERROR` on `stop`
- [x] handle only `type === component` webhook messages

Preview using example error from https://docs.appmixer.com/appmixer/getting-started/system-webhooks#example-data-validation-error
image


src/appmixer/system/core/OnAnyFlowComponentError/OnAnyFlowComponentError.js
```diff
@@ -0,0 +1,70 @@
...
+
+    async stop(context) {
+
+        // Get current WEBHOOK_FLOW_COMPONENT_ERROR from system config
+        const config = await context.callAppmixer({
+            endPoint: '/config',
+            method: 'GET'
+        });
+        const urls = config.find(c => c.key === 'WEBHOOK_FLOW_COMPONENT_ERROR')?.value?.split(',');
+        const newUrl = context.getWebhookUrl();
```
martin-krcmar, line 37: Does this work? You don't have the `{ enqueueOnly }` here, so the URL will be different than the one in the `start` and the the filter will not find it, because it does string comparison. So my guess is that the URL will not be unregistered, when the flow is stopped and will stay there forever.

src/appmixer/system/core/OnAnyFlowComponentError/OnAnyFlowComponentError.js
```diff
@@ -0,0 +1,70 @@
...
+            method: 'GET'
+        });
+        const urls = config.find(c => c.key === 'WEBHOOK_FLOW_COMPONENT_ERROR')?.value?.split(',') || [];
+        // Use enqueueOnly to automatically enqueue the webhook in the `receive` method.
+        const newUrl = context.getWebhookUrl({ enqueueOnly: true });
+        const updatedUrls = [...urls, newUrl].join(',');
+
+        await context.log({ step: 'Update config', newUrls: updatedUrls, oldUrls: urls });
+        // Update WEBHOOK_FLOW_COMPONENT_ERROR in system config
+        await context.callAppmixer({
```
martin-krcmar, line 19: There is a slight chance of messing things up. If two of these components are started simultaneously, one URL could get lost. That should not happen, since this is meant for admins only. And it does not make much sense to have more than 1 in the system. But using `locks` would make it safer. 
But even with a lock, there is a small chance, that the admin changes this in the Backoffice at the same time ...

martin-krcmar (changes requested):

src/appmixer/system/core/OnAnyFlowComponentError/OnAnyFlowComponentError.js
jirihofman: Fixed

src/appmixer/system/core/OnAnyFlowComponentError/OnAnyFlowComponentError.js
jirihofman: I can imagine a case where admin is checking for different error types - for example per connector. Thus using this component multiple times.

I added a lock named `OnAnyFlowComponentError` for both stop and start functions. Could we use this new lock created by the component also in the Backoffice?

vtalas (approved):

martin-krcmar (approved):

PR #139, opened 2024-07-31
added support for private keys in the ftp component

lucvankessel:
We were in need of a key option in the FTP component.
This is the implementation i came up with.
I have tested it  that it still works with password and the new key option.
Let me know if there are things you would like to see different or changed.

### DIFF
--- a/src/appmixer/utils/ftp/auth.js
+++ b/src/appmixer/utils/ftp/auth.js
@@ -31,7 +31,7 @@ module.exports = {
             privatekey: {
                 type: 'textarea',
                 name: 'PrivateKey',
-                tooltip: 'Private key; either use this or the password field, this only works with sFTP. PEM is the only supported format for keys that can be used.'
+                tooltip: 'Private key; either use this or the password field. This only works with SFTP. PEM is the only supported format for keys that can be used.'
             },
             secure: {
                 type: 'text',
--- a/src/appmixer/utils/ftp/auth.js
+++ b/src/appmixer/utils/ftp/auth.js
@@ -16,7 +16,7 @@ module.exports = {
             port: {
                 type: 'text',
                 name: 'Port',
-                tooltip: 'Default 21.'
+                tooltip: 'Default 21. For SFTP the default is 22.'
             },
             username: {
                 type: 'text',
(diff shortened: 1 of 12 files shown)


src/appmixer/utils/ftp/client.js
```diff
@@ -79,7 +85,23 @@ class FtpClient {
      */
     static isFtp(secure) {
 
-        return ['yes', 'implicit', ''].includes(secure);
+        return ['yes', 'implicit'].includes(secure);
```
lucvankessel, line 88: the `''` the default to be the secure option

jirihofman:
Thank you for your contribution @lucvankessel 🎉 . We will review the change some time next week, even though it looks good at first glance.

src/appmixer/utils/ftp/client.js
```diff
@@ -14,10 +15,15 @@ class FtpClient {
     constructor(secure, config) {
 
         this.config = config;
-        this.isFtp = FtpClient.isFtp(secure);
-        if (this.isFtp) {
+
+        if (FtpClient.isFtp(secure)) {
+            if (config.privateKey != '') {
+                throw new Error('Cant use the secure option in combination with a private key');
```
jirihofman, line 21:
```suggestion
                throw new Error('Private key can be used only for SFTP');
```

src/appmixer/utils/ftp/auth.js
```diff
@@ -29,6 +28,11 @@ module.exports = {
                 name: 'Password',
                 tooltip: 'Password.'
             },
+            privatekey: {
+                type: 'textarea',
+                name: 'PrivateKey',
+                tooltip: 'Private key; either use this or the password field, this only works with sFTP. PEM is the only supported format for keys that can be used.'
```
jirihofman, line 34:
```suggestion
                tooltip: 'Private key; either use this or the password field. This only works with SFTP. PEM is the only supported format for keys that can be used.'
```

jirihofman (changes requested):
Minor wording update

src/appmixer/utils/ftp/auth.js
lucvankessel: @jirihofman implemeted both changes

jirihofman (approved):
Great, thank you @lucvankessel

I just bumped the version and added changelog entry

PR #147, opened 2024-08-08
Clickup; fixes, Updated CreateTask and FindTasks with dynamically loaded…

DavidDurman:
… spaces and lists to select from. New components GetSpaces and GetLists. Fix FindTasks one-by-one output type.

NEW:

image

Fixed:

Space ID and List ID selections added to CreateTask, FindTasks:

image


src/appmixer/clickup/bundle.json
```diff
@@ -1,7 +1,8 @@
 {
     "name": "appmixer.clickup",
-    "version": "1.0.1",
+    "version": "2.0.0",
     "changelog": {
+        "2.0.0": ["Updated CreateTask and FindTasks with dynamically loaded spaces and lists to select from. New components GetSpaces and GetLists."],
```
vtalas, line 5: @DavidDurman  please describe the breaking change and outline the steps for migration. also please move the message to the end of the list

vtalas (changes requested):
FindTasks should have the `notFound` output port for cases when no results are found

vtalas (approved):

PR #156, opened 2024-08-12
Slack: add Direct messaging

jirihofman:
Issue: https://github.com/clientIO/appmixer-components/issues/1812
- `ListUsers`
- `SendPrivateMessage`

### DIFF
--- a/src/appmixer/slack/bundle.json
+++ b/src/appmixer/slack/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.slack",
-    "version": "2.2.3",
+    "version": "2.3.0",
     "engine": ">=6.0.0",
     "changelog": {
         "1.0.1": [
@@ -14,6 +14,9 @@
         ],
         "2.2.3": [
             "Add new components: UpdateMessage, DeleteMessage, UploadFile, DeleteFile and GetFile."
+        ],
+        "2.3.0": [
+            "Add new components: SendDirectMessage."
         ]
     }
 }
--- a/src/appmixer/slack/bundle.json
+++ b/src/appmixer/slack/bundle.json
@@ -16,7 +16,7 @@
             "Add new components: UpdateMessage, DeleteMessage, UploadFile, DeleteFile and GetFile."
         ],
         "2.3.0": [
-            "Add new components: SendDirectMessage."
+            "Add new components: SendDirectMessage and ListUsers."
         ]
     }
 }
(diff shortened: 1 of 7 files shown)


vtalas:
- [ ] add the scheme property for the File Id output type (same as here https://github.com/clientIO/appmixer-connectors/blob/0aa9f7b99eeb5242c49fc7d5327b3d26f34c0981/src/appmixer/google/drive/UploadFile/component.json#L91)

src/appmixer/slack/list/ListUsers/ListUsers.js
```diff
@@ -0,0 +1,179 @@
...
+                method: 'GET',
+                headers: {
+                    'Authorization': `Bearer ${context.auth.accessToken}`
+                },
+                url: 'https://slack.com/api/users.list',
+                params: { limit }
+            });
+
+            if (!data) {
+                throw new context.CancelError(response.statusText);
```
vtalas, line 45: is this correct? won't we lose the re-try queue on network errors?

vtalas: `isSource` should ignore all errors and returns an empty list when error occurs, correct?

src/appmixer/slack/messages/SendDirectMessage/component.json
```diff
@@ -0,0 +1,110 @@
...
+                        "type": "select",
+                        "label": "User",
+                        "tooltip": "Select the user to send the message to.",
+                        "index": 1,
+                        "source": {
+                            "url": "/component/appmixer/slack/list/ListUsers?outPort=out",
+                            "data": {
+                                "transform": "./transformers#usersToSelectArray",
+                                "messages": {
+                                    "in/outputType": "array",
```
vtalas, line 64: wouldn't be nice to have there a limit for the dynamic input ?

src/appmixer/slack/bundle.json
```diff
@@ -14,6 +14,9 @@
         ],
         "2.2.3": [
             "Add new components: UpdateMessage, DeleteMessage, UploadFile, DeleteFile and GetFile."
+        ],
+        "2.3.0": [
+            "Add new components: SendDirectMessage."
```
vtalas, line 19: SendDirectMessage and ListUsers

src/appmixer/slack/list/ListUsers/ListUsers.js
jirihofman: I now changed it to throw `Error` so that it can automatically retry. The original code was inspired by `makeRequest` from `slack-commons.js` which is used for all other Slack API calls. This may need refactoring as well.

vtalas (approved):

PR #157, opened 2024-08-15
CreateSpeech, CreateTranscription, Moderate, TransformTextToJSON

DavidDurman:


src/appmixer/utils/ai/CreateSpeech/CreateSpeech.js
```diff
@@ -0,0 +1,33 @@
...
+
+module.exports = {
+
+    receive: async function(context) {
+
+        const { input, voice, responseFormat, speed } = context.messages.in.content;
+        const apiKey = context.config.apiKey;
+
+        if (!apiKey) {
+            return new context.CancelError('Missing \'apiKey\' system setting of the appmixer.utils.ai module pointing to OpenAI. Please provide it in the Connector Configuration section of the Appmixer Backoffice.');
```
vtalas, line 11: @DavidDurman there must be `throw` instead of the `return`

src/appmixer/utils/ai/GenerateImage/GenerateImage.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
         if (imageUrl) {
             const response = await context.httpRequest.get(imageUrl, { responseType: 'stream' });
             const readStream = response.data;
-            const filename = `generated-image-${new Date}.png`;
+            const filename = `generated-image-${(new Date).toISOString()}.png`;
             const file = await context.saveFileStream(filename, readStream);
             return context.sendJson({ fileId: file.fileId, prompt, size }, 'out');
```
vtalas, line 37: - wouldn't be a good idea to send also the image url to the output? 
 - `if (imageUrl) { ` is there a change there is no `image` in the response? shouldn't we throw an error in such case?

DavidDurman: I wouldn't send imageUrl since that resides OpenAI while these utilities are opaque when it comes to the service behind them.

src/appmixer/utils/ai/DescribeImages/DescribeImages.js
```diff
@@ -0,0 +1,79 @@
...
+        const { data } = await context.httpRequest.post(url, payload, {
+            headers: {
+                'Authorization': `Bearer ${apiKey}`,
+                'Content-Type': 'application/json'
+            }
+        });
+
+        let answer = '';
+
+        if (data && data.choices) {
```
vtalas, line 73: if there is no answer (for any reason) it will respond with the '' as success response. I think we should throw an error instead if the `choices` are not present

DavidDurman: Are you sure?

DavidDurman: I don't think this can ever happen though: https://platform.openai.com/docs/api-reference/chat/object

vtalas (approved):

PR #162, opened 2024-08-22
utils.controls.Digest

DavidDurman:
image

### DIFF
--- a/src/appmixer/utils/controls/Digest/Digest.js
+++ b/src/appmixer/utils/controls/Digest/Digest.js
@@ -1,0 +1,122 @@
+'use strict';
+
+const parser = require('cron-parser');
+const moment = require('moment');
+
+module.exports = {
+
+    async receive(context) {
+
+        const { threshold, generateOutputPortOptions, outputType = 'array' } = context.properties;
+        if (generateOutputPortOptions) {
+            return this.getOutputPortOptions(context, outputType);
+        }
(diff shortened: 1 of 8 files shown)


vtalas:
test: 

flow https://my.qa.appmixer.com/designer/f41f62c9-f4d3-4fa2-bab3-62ecb5cc2cb5/7f7ffa5c-a36d-4755-86e1-39ccb8b258b7
image

input
image


flow interates through the input array values, Digest has the threshold set to 2, output type: one at a time


results:
values `11111...` and `2222...` get duplicated: 
image

The `lock` should be implemented to prevent this behavior

vtalas:
the `index` on the output: 
image

It's the index within the threshold, shouldn't be better to implement the `total index` instead?

DavidDurman:
Lock: implemented.
@vtalas I don't see how "total index" would be useful? The index is just the standard index we have for array-type of outputs.

vtalas (approved):

RamBachkheti:
Issue: Getting "Invalid Characters, got value: undefined" while starting the flow when valid values for "Minute" and "Hour" text values are set, please see the following image:
- 
Screenshot 2024-10-28 at 12 00 18 PM
Screenshot 2024-10-28 at 11 59 41 AM

DavidDurman:
Fixed in https://github.com/clientIO/appmixer-connectors/pull/288

PR #173, opened 2024-09-04
rabbitmq (new)

DavidDurman:


vtalas:
- [ ] - Publish - when the exchange is defined it throws an error when starting the flow 
image
The `Exchange` input is not required, therefore I assuming the default exchange is selected, correct ?

- [ ] - SendToQueue: throws an error when I send a message to the webhook: 
image
image

it's  not always happening, sometimes it succeeds: 
image
However, even when the SendToQueue succeeds, the NewMessage is not triggered

DavidDurman:
No, when the Exchange is empty in Publish (undefined or empty string) - a special case - it will send directly to the queue named by the routing key. That's how the amqplib behaves: https://amqp-node.github.io/amqplib/channel_api.html#channel_publish

vtalas (approved):

RamBachkheti:
Issue 1: [SendToQueue] Providing header values showing validation message. I am following the same format mentioned in tooltip, please see the following image:
-  Screenshot 2024-10-21 at 11 24 57 AM

Thanks!

vtalas:
fix in https://github.com/clientIO/appmixer-connectors/pull/241

@RamBachkheti  ready for re-test 1.0.1

RamBachkheti:
@vtalas: Thanks for sharing the RabbitMQ admin panel details. I am currently testing the "Publish" and "SendToQueue" actions, but I'm unsure how to verify the actual results against the expected ones. Could you please take a look at the following video and guide me on how or where to verify the actual results? I haven't been able to find any options in the RabbitMQ admin panel to check the published messages or those sent to the queue.

https://github.com/user-attachments/assets/9fe778dd-57d8-4db6-89b6-7f0b215bd3f7

Test Flow: https://my.stg.appmixer.com/designer/3398bc69-84b7-43b6-8c0b-3e074f8c1853/2cc21e14-aa43-44aa-848a-836d624621f6

Thanks!

vtalas:
here are some basic essentials to get started: 

### Queues
- **Purpose:** A queue is where messages are stored until they are consumed by a consumer. Each message in a queue is typically received by one consumer only.
- **Behavior:** Queues hold messages and ensure that they are delivered in the order they were received (FIFO: First In, First Out), unless specific configurations like priorities are set.
- **Durability:** Queues can be durable, ensuring that they survive a broker restart, provided they are explicitly declared as durable.
- **Binding:** Queues are bound to exchanges which determine the routing of messages to these queues. You do not publish messages directly to a queue; messages are published to an exchange and routed to the appropriate queue(s) based on bindings.
- **Message Acknowledgments:** Consumers must acknowledge receiving a message for it to be removed from the queue, ensuring reliable delivery.

### Exchanges
- **Purpose:** Exchanges take messages from producers and route them to one or more queues based on rules defined by the exchange type and bound queue(s).
- **Types:** There are different types of exchanges which determine the routing logic:
  - **Direct Exchange:** Routes messages with a specific routing key to queues that are bound with a matching key.
  - **Fanout Exchange:** Routes messages to all bound queues indiscriminately, irrespective of routing keys.
  - **Topic Exchange:** Routes messages to queues based on matching between a message's routing key and the pattern specified in the binding.
  - **Headers Exchange:** Routes messages based on header values instead of the routing key.
- **No Message Storage:** Exchanges do not store messages. They just route messages to queues.
- **Bindings and Routing:** You define a binding between an exchange and a queue with specific routing rules. The exchange uses these bindings to determine how to route messages to the appropriate queues.
- **Durability:** Like queues, exchanges can also be made durable to ensure they survive broker restarts.


@RamBachkheti you can see the message in the rabbit mq management, as reading a message is destructive action - once it's read, it's lost. 

I'd suggest a flow with NewMessage, Publish, SendToQueue and Webhooks, configured for a single queue . Data sent to one of the webhooks should be consumed by the NewMessage trigger. Also you can try  sending a message to a queue from the management studio
image


image

image

image

IMPORTANT: create your own queue, you can't be sure that the existing queue is not consumed in any flow anywhere else

vtalas:
@RamBachkheti  version 1.0.2 is ready for testing

RamBachkheti:
Verified the fix on staging.

PR #177, opened 2024-09-11
Google.gmail Auth.js + Find Emails UI fix

sayam-nasir:
Replaced the usage of Googleapis in auth.js with axios and endpoint calls

Added conditional property in FindEmails inspector to not show limit input if output type is First

Fixed typo in package.json


src/appmixer/google/gmail/package-lock.json
jirihofman, line 1: We currently don't support package-lock files. Please remove it.

src/appmixer/google/gmail/package.json
```diff
@@ -1,7 +1,8 @@
 {
-    "name": "appmixer.goog.gmail",
+    "name": "appmixer.google.gmail",
     "version": "1.0.0",
     "dependencies": {
+        "axios": "^1.7.7",
```
jirihofman, line 5: Use `context.httpRequest` instead of `axios`. This way you don't need to install any dependencies. `context.httpRequest` is the same as `axios`.

jirihofman (changes requested):

src/appmixer/google/gmail/package.json
sayam-nasir: @jirihofman thank you! I have made the changes

jirihofman (approved):

PR #181, opened 2024-09-12
zendesk.tickets (major update), zendesk.chats (removed)

DavidDurman:
- Old Zendesk tickets connector was not working.
- This is a major upgrade with updates to non-working components together with getting on-par with Zapier when it comes to triggers/actions.

Q: It's been under heavy discussions whether to move this connector under the `zendesk/` directory or leave it as is in `zendesktickets/`.

This PR also updates the `zendesk.chats` connector and fixes https://github.com/clientIO/appmixer-components/issues/1873


jirihofman (approved):

src/appmixer/zendesktickets/package-lock.json
jirihofman, line 1: We don't use `package-lock.json` files yet. Please remove it.

jirihofman (changes requested):

jirihofman (approved):

vtalas (commented):
@DavidDurman  quotas are missing, there should be at least some quotas set

DavidDurman:
The rate limits of Zendesk are very complicated and also depend on the account tier. I'm not sure how this can be implemented using our current quotas mechanism. See https://developer.zendesk.com/api-reference/introduction/rate-limits/#zendesk-suite-plan-limits

vtalas:
- [ ] add at least some basic quotas

vtalas (changes requested):
apply quotas in compoent.json files

vtalas (approved):

RamBachkheti:
@vtalas: The backoffice displays four different tickets labeled 'Zendesk.' Could you please let me know which one I should use? Additionally, could you remove any duplicates to avoid confusion?

Screenshot 2024-10-21 at 6 41 13 AM

vtalas:
@RamBachkheti  it't this one: 
image

we're going to remove deprecated version from the marketplace

RamBachkheti:
Issue 1: [CreateUser] Getting "error:DatabaseConflict", please see the following image:
- Screenshot 2024-10-22 at 11 27 33 AM
<hr>

Issue 2: [UpdateOrganization] "Domain Names" should have validation message when domain formate doesn't match, please see the following image:
Screenshot 2024-10-22 at 11 26 22 AM

DavidDurman:
@vtalas why is Zendesk requiring me to enter clientID/secret on QA? This wasn't the case before and the standard OAuth was working fine:

image

vtalas:
> @vtalas why is Zendesk requiring me to enter clientID/secret on QA? This wasn't the case before and the standard OAuth was working fine:
> 
> image

@DavidDurman  it's due to bug on in the authub (https://github.com/clientIO/appmixer-core/issues/2640), when the `globasKeys` is set to `true`, it won't ask for the clientId & secret.

DavidDurman:
@RamBachkheti do you have repro steps to Issue nr. 1? It's working fine for me with these values:

image

vtalas:
> Issue 2: I don't think we have a way to define a validation for "comma separated list of domains". What do you think @vtalas .

Surely we have, you can use the `pattern`  (regex). On the other hand, the velation message is not pretty in that case 
image

Also even it's great to catch as many errors as possible at the design phase I don't think it's crucial to introduce a regex in this case. @DavidDurman  you decide. I would rather consider including this note  
image

https://developer.zendesk.com/api-reference/ticketing/organizations/organizations/#update-organization

PR #182, opened 2024-09-13
google.spreadsheets.NewRow (major) introduce index column 

sayam-nasir:


vtalas:
this looks promising! here are few notes: 
 - initialization of the "index column" should be done in the `start` method. See [here ](https://github.com/clientIO/appmixer-connectors/blob/da5501a7d06d63794acf6f12a834425b64803538/src/appmixer/google/drive/NewFileOrFolder/NewFileOrFolder.js#L104-L103) where we get the page token first.

- the change as it is right now is a breaking change - there is a new required field 
- there is an issue when you add a blank row in between - it detect the last row as a new 
image

vtalas:
It's a significant improvement; however, there are still some limitations and edge cases. Tracking new rows is not an easy task. Perhaps we can define an explicit ruleset for when NewRow functions properly. These rules will be included in the component description.

vtalas (approved):

vtalas (approved):

PR #184, opened 2024-09-17
Google.Gmail Adding Index and Count in Outputs

sayam-nasir:
Added Index and Count in Output port options of FindEmails


jirihofman (approved):

src/appmixer/google/gmail/FindEmails/component.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "appmixer.google.gmail.FindEmails",
     "author": "Appmixer <info@appmixer.com>",
-    "description": "Finds emails in Gmail based on a search query. You can retrieve emails as a single item, multiple items, or save them directly to a file. Each message resource contains only the Email ID and Thread ID. To get detailed information about specific emails, use the Get Email component. For guidance on search queries, refer to the <a href=\"https://support.google.com/mail/answer/7190?hl=en\" target=\"_blank\">Gmail Search Documentation</a>.",
+    "description": "Finds emails in Gmail based on a search query. You can retrieve emails as a single item, multiple items, or save them directly to a file. Each message resource contains only the Email ID and Thread ID. To get detailed information about specific emails, use the Get Email component. For guidance on search queries, refer to the <a href=\"https://support.google.com/mail/answer/7190?hl=en\" target=\"_blank\">Gmail Search Documentation</a>. This Component will return 500 records at Maximum",
```
vtalas, line 4: instead of the "This Component will return 500 records at Maximum" 

"This component can return a maximum of 500 records"  or "This component has an upper limit of 500 records it can return"

not sure abut the capital C an M

vtalas (changes requested):
I also noticed the `index` and `count` is missing in the ListEmail (even though it's private, I think we should be consistent)

- GetAttachment is still having the dynamic input (20 most recent emails) this  should be removed

sayam-nasir:
> I also noticed the `index` and `count` is missing in the ListEmail (even though it's private, I think we should be consistent)
You are right, I left it out because the component is now private and we are using ListEmails for the dropdown input in GetAttachments as well
> * GetAttachment is still having the dynamic input (20 most recent emails) this  should be removed
Yes, I left it there because of the unique functionality of the GetAttachments action.
Get Attachments work in the following manner 
Select an email which has attachments (or insert the email ID of an email which has attachments)
Then
The attachment input option will show up and it will have all attachments of that email selected in step 1
Then user selects the attachment to download

It is not possible that after inserting email ID from another step, the user selects the attachments from the dropdown while the flow is in running state. So I have kept it as it is as this component  can be run individually to save attachments from Gmail to Appmixer

@vtalas Let me know how you think we should proceed with this

vtalas (approved):

PR #188, opened 2024-09-20
hubspot (major update) webhook improvements (engine v5+)

jirihofman:
Issue: https://github.com/clientIO/appmixer-components/issues/1700

## Done
- [x] Use `portalId` when tracking subscribed components internally.
- [x] Remove `bluebird` promises
- [x] Remove concurrency 100 - HubSpot doesn't send more than 100 events in a payload.

## Logs in NewRelic
These are the logs with default log level to `info`
image


src/appmixer/hubspot/bundle.json
```diff
@@ -42,6 +42,9 @@
         "2.2.0": [
             "New component: Engagements - GetEmails: Read a batch of emails by internal ID, or unique property values.",
             "Update component: GetContact: fix support for adding properties, add support for retrieving associations."
+        ],
+        "3.0.0": [
+            "(breaking change) Improve webhook handling by better event distribution. Affected components: NewContact, UpdatedContact, NewDeal, UpdatedDeal."
```
jirihofman, line 47: @vtalas It is breaking because all the trigger components need to be stopped and started again in order to use the keys in `context.store` with `portalId`.

src/appmixer/hubspot/routes.js
```diff
@@ -116,35 +116,48 @@ module.exports = async (context) => {
             auth: false,
             handler: async (req) => {
 
+                await context.log('info', 'hubspot-plugin-route-webhook-hit', { payload: req.payload });
```
martin-krcmar, line 119: Be careful with logs. For development this is fine. However, logging every incoming payload in production can cause issues. If the payload is big and there are lots of incoming requests, that could take lots of disk space.

src/appmixer/hubspot/bundle.json
martin-krcmar: One thing. We want to rewrite this to use the new `events/listeners`. 
We discussed with @vtalas that we want to fix it first for version 5, ideally without a breaking change. But that may not be possible. If it is not possible to do it without the breaking change, then I would not release this, until we have the version with `events/listeners`. The reason:
- A customer (even a customer already on version 6) will update the Hubspot, and all the flows have to be started again. 
- Then, in a week we will release another breaking version that works with the `events/listener` and the customer will have to install a new major version again. 

So either write directly a version with `events/listeners`, or prepare both and release `3.0.0` without dependency on Appmixer >=6 and 3.0.1 with dependency on Appmixer >= 6

src/appmixer/hubspot/routes.js
jirihofman: I put potentially large logs into `trace` level. Also changed the messages (2nd argument) a bit to make it more searchable.

src/appmixer/hubspot/bundle.json
jirihofman: @martin-krcmar This is a fix for Appmixer <6 only (no `events/listeners` support). It is a breaking change as mentioned above.

What should be the next steps then? I can imagine something like this
- Hubspot 3.0.0 (this PR) merged into `dev`
- Hubspot 4.0.0 (another PR) with this:
    - rewrite 3.0.0 to `events/listeners`
    - for Appmixer >=6 only
    - merge into `dev`
- Release both versions at the same time

vtalas (approved):

vtalas:
@RamBachkheti v3.0.0 is ready for testing

PR #199, opened 2024-09-26
Airtable: fix outputPortOptions

jirihofman:
Fix for https://github.com/clientIO/appmixer-components/issues/1909.

- ListBases, ListTables
- has to be a breaking change as there is a change in outputPorts


sayam-nasir (approved):

vtalas (approved):
It's unfortunate that this needs to be a breaking change; however, avoiding it by changing the `result` property to `array` in the sendArrayOutput method would break the ListRecords action.

PR #200, opened 2024-09-26
Asana. Fix for project and task dropdown in triggers

sayam-nasir:


vtalas (changes requested):
update the bundle.json as well please

vtalas (approved):

PR #205, opened 2024-10-03
hubspot (major update) webhook improvements (engine v6+)

jirihofman:
Followup to #188. This is a fix for Appmixer v6+ where `events` and `listeners` are available.

Compare with v3+, keep its functionality. Just use different trigger subscription in Appmixer

### Changed
- [x] using `context.addListener` instead of `context.service.stateAddToSet`
- [x] using `context.removeListener` instead of `context.service.stateRemoveFromSet`
- [x] remove `context.setTimeout` - the component is no longer split into `webhook` and `timeout` parts. This "optimization" is no longer needed as we reduced the number of calls to this component by using `portalId` and Appmixer `listener`
- [x] using batch endpoint to prevent 429 when doing bulk import of contacts https://github.com/clientIO/appmixer-components/issues/1942
- [x] using `context.triggerListeners` instead of `context.triggerComponent`
- [x] call "subscribe" in HubSpot from AH, not tenant. Because it needs `appId` available only in AH
- [x] don't call "unsubscribe" in HubSpot for now. There is no easy way to sync subscriptions across multiple tenants. Idea:
    - do it in a job in AH, hourly, 4 subscription types (2x contact, 2x deal)
    - using new lock named `hubspot-subscribe`
    - this lock is also used when getting subscriptions
- [x] imports in `routes.js` need to be from "root" files as AuthHub doesn't load component files

### Fixed
- [x] Race condition when subscribing to events on flow start
    - a flow has two components: `NewContact`, `UpdatedContact`
    - only the last one is subscribed
- [x] Sometimes `UpdatedContact` and `UpdatedDeal` events can be ignored. If multiple properties were changed, but the one that is not watched is passed to trigger the component.
- [x] Ignore update events sent for the same object within 5 seconds
- [x] compare `createdAt` vs `updatedAt` for `Updated*` triggers

### Optimized
- [x] `UpdatedContact` - subscribing only for property changes that are in `CreateContact` inspector.
- [x] `UpdatedDeal` - subscribing only for property changes that are in `CreateDeal` inspector.
- [x] batch requests for contact/deal details, similar to [Xero webhook handler](https://github.com/clientIO/appmixer-connectors/blob/dev/src/appmixer/xero/commons.js#L47).
- [x] filter unnecessary events in AH plugin to reduce number of times a component is triggered but does nothing

Issue https://github.com/clientIO/appmixer-components/issues/1700


jirihofman:
This will be ready for review when v3.0.0 is tested - #188

vtalas (approved):

PR #206, opened 2024-10-03
JIRA.Components (minor) description updates

sayam-nasir:
Updating description verbs from Triggered to Get/Update/Create so that there is no confusion between actions and triggers

### DIFF
--- a/src/appmixer/jira/issues/GetIssue/component.json
+++ b/src/appmixer/jira/issues/GetIssue/component.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.jira.issues.GetIssue",
-    "description": "When triggered, retrieves an issue.",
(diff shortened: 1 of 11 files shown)


src/appmixer/jira/issues/GetIssue/component.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.jira.issues.GetIssue",
-    "description": "When triggered, retrieves an issue.",
+    "description": "Gets a selected Issue using its ID..",
```
jirihofman, line 3:
```suggestion
    "description": "Get a selected Issue using its ID..",
```

To make it consistent with the rest.

vtalas (approved):

PR #208, opened 2024-10-04
asana.CreateProject (patch update) Color Palette not showing colors

sayam-nasir:
Problem 1
The initial issue was that the color palette in the Asana Create Project component displayed Asana’s color names (e.g., "dark-pink") but these names were not valid CSS color values. This caused the color picker in the UI to not display colors correctly, as it could not interpret non-standard CSS values like "dark-pink".

Problem 2
some components have wrong description that can cause confusion between action and trigger

Fix:
Updated the color palette to use valid hex codes for the color options instead of Asana’s internal color names. The color names were still required by the Asana API, so I introduced a mapping between hex values and Asana color names. The UI now displays hex values, and when submitting the project, the hex value is mapped back to the corresponding Asana color name before being sent to the API. This ensures both proper color display in the UI and compatibility with Asana’s API.

Updated the verbs of the descriptions of action components


src/appmixer/asana/projects/CreateProject/component.json
```diff
@@ -204,26 +31,36 @@
         {
             "name": "newProject",
             "options": [
```
vtalas, line 33: @sayam-nasir  you've removed some of the output properties e.g `archived`, `notes`, `workspace`.... and some more are gone. Is there a reason for that? if so, it's a breaking change then
Also, could you please update the labels to match our guiidelines (Title Case format)?

src/appmixer/asana/projects/CreateProject/CreateProject.js
```diff
@@ -1,53 +1,120 @@
...
+            dueDate: {
+                label: 'Due Date',
+                type: 'date-time',
+                tooltip: 'Select the project due date (or date-time).',
+                index: 5
+            },
+            color: {
+                label: 'Color',
+                type: 'color-palette',
+                options: Object.entries(asanaColorMap).map(([asanaColorName, hexValue]) => ({
```
vtalas, line 90: the inspector is generated dynamically due to this color palette inputs, is that right? 
is it really needed? it's only converting hard-coded `asanaColorMap` to another format (key: value -> value:key) why it's not in the `{value: hex, content: "name"}` directly in the component.json? 

it could be reasonable if the `asanaColorMap` is somehow configurable. 

BTW you don't need to make the whole inspector dynamic because of one input, you can generate inspector field separately, like it's here: https://github.com/clientIO/appmixer-connectors/blob/73c79f49274fc8f602c267de2e4d8dab27096424/src/appmixer/utils/forms/FormTrigger/component.json#L22

vtalas (changes requested):

src/appmixer/asana/projects/CreateProject/component.json
sayam-nasir: @vtalas this has beem fixed, we dont need a change in this

src/appmixer/asana/projects/CreateProject/CreateProject.js
sayam-nasir: @vtalas youre right, 
I will just take the input from the user, keep inspector in component.json. Create a mapper in CreateProject.js that will map user's input to asana specific color names and send to the API that will be cleaner simpler solution

vtalas (approved):

PR #215, opened 2024-10-08
Slack.SendChannelMessage, SendPrivateChannelMessage (patch update) Bug fixes

ZbynekPelunek:
Fix for issues with **SendChannelMessage**, **SendPrivateChannelMessage** channels to having the option to use outports from previous components in a flow.
Updated labels in some components where it was just lowercase.

For the **NewUser** issue however, we would need for the API to send the newest created user first but I haven't found a way how to make that happen yet. I don't even see a created date in the API response that I could work with. Need to investigate this more.

[Link to the issue](https://github.com/orgs/clientIO/projects/9/views/17?pane=issue&itemId=81841706)


ZbynekPelunek:
This PR is only for the Issue2, the NewUser issue will be fixed later by making the component into webhook

src/appmixer/slack/bundle.json
```diff
@@ -18,6 +18,11 @@
         "2.4.1": [
             "Add new components: SendDirectMessage and ListUsers.",
             "Add new components: AddReaction, DeleteReaction, ListReactions and GetMessageReactions."
+        ],
+        "3.0.0": [
+            "(breaking change) SendChannelMessage, SendPrivateChannelMessage: in component.json moved channelId from properties to inPort.",
```
sayam-nasir, line 23: @ZbynekPelunek are we sure that this is a breaking change? @vtalas to confirm

ZbynekPelunek:
> @ZbynekPelunek I have reviewed the code and tested it on QA it looks fine to me however there is one observations that came along while testing which are not necessarily related to the code changes made by you
> 
> The Create Channel and Create Private Channel have a pattern set for channel names "pattern": "^[a-z0-9-_]{1,80}$" but there is no tooltip to explain that it wont accept capital letters or spaces etc. so we can update the tooltip text to just show this in a better way

Actually, there is tooltip for that in both:
Screenshot 2024-10-11 151824

It does mention the lowercase letters but nothing about spaces, however what I find odd is that this tooltip says that the maximum length of the name is 21 letters, but in the code it is set to 80.
According to Slack API, the correct one is 80: https://api.slack.com/methods/conversations.create , any idea why the tooltip says 21?

sayam-nasir:
@ZbynekPelunek yes I saw that tooltip but it should somehow specify that it wont work if a capital letter is used etc. I think by using the word "must" before "include"

also, seems like 21 chars is a mistake because the regex is for 80 and it works for 80 chars (verified) so maybe update it to somethinng like 

"Name of your channel. It can contain up to 80 characters and must include only lowercase letters, numbers, underscores, and dashes."

ZbynekPelunek:
@sayam-nasir I played a little bit with the tooltip, here are my other versions of how it can sound:
Screenshot 2024-10-11 171559
The first one is your proposition.
The second and third my ideas of how it could sound.
The fourth one is taken directly from the API documentation.

Which one do you think sounds the best? Or we can make any combination of these.

sayam-nasir:
@ZbynekPelunek all are good. I would suggest to go with the third option

src/appmixer/slack/bundle.json
vtalas: we moved the input from properties to input message, therefore it's a breaking change

sayam-nasir (approved):

PR #218, opened 2024-10-09
Google.Bigquery.(patch update).Fix for error "Cannot read properties of undefined (reading 'id')"

sayam-nasir:
In the Triggers NewRow, UpdatedRow and DeletedRow
The SQL Query needs project ID. which was being fetched using a function getProjectId. In this function we were trying to find the project ID out of a list of all projects as whichever projects numericid would match the first part of clientID, we would take that project's projectID.

This approach was not working as none of the projects' numeric ID was matching anything in the list of projects

Hence, now an input field of select type has been introduced in the triggers which helps the user to select one project out of all the projects in their bigQuery playground and then the query is run on this project's ID.

### DIFF
--- a/src/appmixer/google/bigquery/DeletedRow/DeletedRow.js
+++ b/src/appmixer/google/bigquery/DeletedRow/DeletedRow.js
@@ -24,11 +24,9 @@ module.exports = {
 
     async start(context) {
 
-        const { query, idField } = context.properties;
+        const { query, idField, projectId } = context.properties;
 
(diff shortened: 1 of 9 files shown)


jirihofman (changes requested):
I think we should also remove the `getProjectId` function.

sayam-nasir:
removed the function getProjectId because it is not used anywhere anymore

vtalas (changes requested):
projectId  should be marked as `required`

jirihofman (approved):

PR #219, opened 2024-10-09
ServiceNow (patch) remove NewRecord for Appmixer v5

jirihofman:
Adding a dependency on Appmixer v6 because `NewRecord` is using `context.addListener`.

But what about the existing customers on v5 who use `ServiceNow`? Should we offer an update that will remove the connector? Should it be breaking change instead?

### In this PR

- [x] release version 1.1.1 for the Appmixer 5: 
- [x] remove NewRecord component
- [x] no engine dependency in the bundle.json

This will be followed by a PR for Appmixer v6


vtalas:
customers with Appmixer v5 can install ServiceNow 1.1.0 with the NewRecord and it's not working there.
I think we should do this
 - release version 1.1.1 for the Appmixer 5: remove NewRecord component, no engine dependency in the blundle.json 
 - release version 1.1.2 will have NewRecord and de "engine": ">=6.0.0", dependency

vtalas (approved):

PR #220, opened 2024-10-09
Twitter.(major).Rebranding from Twitter to X

sayam-nasir:
Name of the connector changed from Twitter -> X
Icon changed from Twitter -> X
Description updated


vtalas (changes requested):
@sayam-nasir I think it's not necessary moving Twitter to X, causing the breaking change. All we need to do is change the labels and descriptions in service.json and update icons

sayam-nasir:
@vtalas alright, makes sense. I have minimized the changes now

src/appmixer/twitter/bundle.json
```diff
@@ -1,12 +1,15 @@
 {
     "name": "appmixer.twitter",
-    "version": "2.0.2",
+    "version": "3.0.0",
```
vtalas, line 3: it's not a breaking change now, right?

sayam-nasir: I think it shouldn't be 
 Only icons and activity descriptions and label of app was changed

vtalas (approved):

PR #225, opened 2024-10-11
servicenow.CreateRecord (major update) Get the json schema for input properties 

vtalas:


vtalas:
- [x] rename common.js to lib.js 
- [ ] fetch json structure and generate inspector

src/appmixer/servicenow/lib.js
```diff
@@ -0,0 +1,256 @@
...
+
+    const { data } = await callEndpoint(context, options);
+
+    await context.staticCache.set(
+        key,
+        data,
+        context.config.listBasesCacheTTL || (2 * 60 * 1000) // 120s
+    );
+
+    return { data };
```
jirihofman, line 41: I think this block should also be inside a `context.lock(key)`. Imagine a flow where different types of eg. issues (table `incident`) are created. Lets say: errors, questions, feature requests. Each `CreateRecord` component will have the same inspector, but different fields (eg repro steps for error, but not for the other two types).

Without lock there will most likely be 3 calls to `callEndpointCached` even though it is cached - because they will fire at the same time.

src/appmixer/servicenow/record/CreateRecord/component.json
```diff
@@ -1,90 +1,55 @@
 {
     "name": "appmixer.servicenow.record.CreateRecord",
     "author": "Jiří Hofman <jirka@client.io>",
-    "description": "Inserts one record in the specified table. Multiple record insertion is not supported by this method.",
+    "description": "Inserts a single record into the specified table. This method does not support inserting multiple records.",
     "version": "1.0.0",
```
jirihofman, line 5:
```suggestion
    "version": "2.0.0",
```

src/appmixer/servicenow/lib.js
```diff
@@ -0,0 +1,275 @@
...
+        const table = tableId ? data?.result : data?.result[0];
+
+        if (table.super_class) {
+            tables = tables.concat(await getTableReferences(context, { tableId: table.super_class.value }));
+        }
+
+        tables.push(table.name);
+        return tables;
+    } catch (e) {
+        console.log(e);
```
jirihofman, line 192:
```suggestion
```

src/appmixer/servicenow/lib.js
```diff
@@ -0,0 +1,275 @@
...
+
+async function callEndpointCached(context, options) {
+
+    let lock;
+    try {
+        lock = await context.lock(context.componentId);
+
+        const key = getCacheKey({ ...options, token: context.auth.username + context.auth.password });
+
+        console.log(key);
```
jirihofman, line 27:
```suggestion
```

src/appmixer/servicenow/lib.js
```diff
@@ -0,0 +1,275 @@
...
+        .createHash('sha256')
+        .update(str)
+        .digest('hex');
+}
+
+async function callEndpointCached(context, options) {
+
+    let lock;
+    try {
+        lock = await context.lock(context.componentId);
```
jirihofman, line 23: Shouldn't the lock name be the same for all `CreateRecord` components with the same options? This will create a different lock for each component. The optimization in reducing API calls will be lost.

jirihofman (changes requested):

jirihofman (approved):

RamBachkheti:
@jirihofman: I am getting "Connecting account failed" error when I try to authenticate with test credentials mentioned in credentials sheet. Due to that I am not able to review this ticket. Can you please check the following video and let me know if I am missing anything here:
-  Uploading Screen Recording 2024-11-13 at 1.47.10 PM.mov…

Thanks!

jirihofman:
@RamBachkheti there is no recording available 🙂 

Did you follow the steps in Service Account sheet for ServiceNow? Usually the instance is hibernated after few days of inactivity. First, you need to "wake it up". After that (5min) the credentials worked for me - `dev265454`.

image

RamBachkheti:
@jirihofman: I have re-uploaded the video, can you please take at look into it and let me know if I am missing anything here

vtalas:
@RamBachkheti  it seems they update the instance url. you are signed as `dev265454` in Appmixer, however screenshot fomr servicenow is from `dev269411` I added an validation step
image

Some service now endpoints are returning random data even the instance is not running.  Getting the schema in the CreateRecord is failing if the instance is not running. Only dev instances hibernate, this should no happen in the prod instances

RamBachkheti:
**Issue 1:[Update Record] The "Sys ID" field is marked as required, even though it already contains the correct value**, please see the following image:
- Screenshot 2024-12-30 at 11 25 03 PM
<hr>

Issue 2:[Get Record] The "Servicenow account/ Table name / Sys ID" field is marked as required, even though it already contains the correct value, please see the following image:
- Screenshot 2024-12-30 at 11 24 48 PM

**Test Flow:**
- https://my.stg.appmixer.com/designer/049dc25c-1b0e-4f70-86a1-a826fb50ba10/d0334efd-be81-445e-8d56-c4456d810bf6

PR #228, opened 2024-10-14
Utils.RSS.NewFeedItem: url is now required fix

ZbynekPelunek:
The URL for RSS feed was not required, it is now.


src/appmixer/utils/rss/NewFeedItem/component.json
```diff
@@ -6,8 +6,14 @@
...
-                "url": { "type": "string", "format": "uri" }
-            }
+                "url": {
+                    "type": "string",
+                    "format": "uri"
+                }
+            },
+            "required": [
+                "url"
+            ]
```
jirihofman, line 16: I think this is a breaking change. Please bump the version in `bundle.json` - version and changelog entry.

src/appmixer/utils/rss/NewFeedItem/component.json
```diff
@@ -24,16 +30,40 @@
...
+                {
+                    "label": "ISO Date",
+                    "value": "isoDate"
+                }
             ]
         }
     ],
     "icon": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiA/PjxzdmcgaGVpZ2h0PSI1MTIiIHZpZXdCb3g9IjAgMCA1MTIgNTEyIiB3aWR0aD0iNTEyIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPjx0aXRsZS8+PHBhdGggZD0iTTEwOC41NiwzNDIuNzhhNjAuMzQsNjAuMzQsMCwxLDAsNjAuNTYsNjAuNDRBNjAuNjMsNjAuNjMsMCwwLDAsMTA4LjU2LDM0Mi43OFoiLz48cGF0aCBkPSJNNDgsMTg2LjY3djg2LjU1YzUyLDAsMTAxLjk0LDE1LjM5LDEzOC42Nyw1Mi4xMXM1Miw4Ni41Niw1MiwxMzguNjdoODYuNjZDMzI1LjMzLDMxMi40NCwxOTkuNjcsMTg2LjY3LDQ4LDE4Ni42N1oiLz48cGF0aCBkPSJNNDgsNDh2ODYuNTZjMTg1LjI1LDAsMzI5LjIyLDE0NC4wOCwzMjkuMjIsMzI5LjQ0SDQ2NEM0NjQsMjM0LjY2LDI3Ny42Nyw0OCw0OCw0OFoiLz48L3N2Zz4="
-}
+}
```
jirihofman, line 69: It seems your IDE trims the last empty line.

jirihofman (changes requested):
~~I think~~ I left some comments below :)

src/appmixer/utils/rss/NewFeedItem/component.json
ZbynekPelunek: Is it though? Since this is a trigger connecter, you wouldn't have anything mapped to it from previous connectors and also if you didn't fill the URL, you would get the error mentioned in the issue. You basically couldn't use the connector with empty URL so it was already technically required.

ZbynekPelunek: Of course I don't understand yet how it works in detail, but if you had an URL filled and I would change it to required, it wouldn't remove the filled URL, right?

src/appmixer/utils/rss/NewFeedItem/component.json
ZbynekPelunek: My apologies, I already made it so my IDE doesn't format on save, so this shouldn't happen in the future

src/appmixer/utils/rss/NewFeedItem/component.json
jirihofman: My bad, it is not a breaking change, just a fix 🤦‍♂️ Sorry for the confusion. This change will definitely not remove any data that has been already entered.

But that still needs a bump (patch version increase) in `bundle.json`. This way customers with the buggy version are offered this fix in their BackOffice.

src/appmixer/utils/rss/NewFeedItem/component.json
ZbynekPelunek: I have put the empty line back in the 1.0.1 commit

jirihofman (approved):
Great, thanks!

PR #233, opened 2024-10-15
virustotal (new)

DavidDurman:
image


vtalas:
@DavidDurman  could you please add a testing api key to the "Service Accounts"?

vtalas (commented):
- please shorten the API display value, to something like `994...xsd`, long API doesn't look good in the designer 
image

vtalas (approved):

PR #242, opened 2024-10-22
Mailchimp.FindSubscriber (minor) adding notFound port in FindSubscribers

sayam-nasir:
for https://github.com/clientIO/appmixer-components/issues/1947#issuecomment-2426116409

### DIFF
--- a/src/appmixer/mailchimp/bundle.json
+++ b/src/appmixer/mailchimp/bundle.json
@@ -1,6 +1,6 @@
 {
   "name": "appmixer.mailchimp",
-  "version": "2.0.5",
+  "version": "2.0.6",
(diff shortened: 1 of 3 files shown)


src/appmixer/mailchimp/bundle.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appmixer.mailchimp",
-  "version": "2.0.5",
+  "version": "2.0.6",
```
vtalas, line 3: @sayam-nasir  we are introducing a new feature - new port, it ought to be "minor" release (2.1.0)

vtalas (approved):

PR #255, opened 2024-10-31
Jotform.NewFormSubmission (patch): removed unnecessary condition

ZbynekPelunek:
The condition is evaluating if the received data has Form ID, I tried to remove it and without it it still works. I also tried to submit a different form that this trigger is not watching and it didn't trigger it, so I don't see a point of it at all. I have running flow with this component and after few hours it didn't trigger with anything unwanted.

### DIFF
--- a/src/appmixer/jotform/core/NewFormSubmission/NewFormSubmission.js
+++ b/src/appmixer/jotform/core/NewFormSubmission/NewFormSubmission.js
@@ -7,7 +7,7 @@ const dependencies = {
 
 module.exports = {
 
-    receive: async function(context) {
+    receive: async function (context) {
 
         if (context.messages.webhook) {
             await context.log({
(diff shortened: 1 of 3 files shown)


src/appmixer/jotform/core/NewFormSubmission/NewFormSubmission.js
```diff
@@ -17,17 +17,20 @@ module.exports = {
...
-            });
-            if (!condition) return null;
+            // const expCondition = dependencies.jsonata('$exists(submissionID)');
+            // context.log({ step: 'expCondition', expCondition });
+            // context.log({ step: 'expCondition out', out });
+            // context.log({ step: 'parameters: context.properties', contextProperties: context.properties });
+            // const condition = await expCondition.evaluate(out, {
+            //     parameters: context.properties
+            // });
+            // if (!condition) return null;
```
jirihofman, line 27: You can remove this code.

Also please see the linting errors.

jirihofman (changes requested):

jirihofman (approved):

PR #256, opened 2024-11-01
Google.Calendar.ListEvents (patch) Bug fixes

sayam-nasir:
A private component ListCalendards had bugs, the path to source endpoint was wrong which was listing calendars to be selected in the inspector


vtalas (approved):
Please note that changing the input key is considered a breaking change. However, in this case, it's a bug fix because the component was not functioning at all.

PR #269, opened 2024-11-06
Jira.UpdatedIssue (patch): fix onTick error

ZbynekPelunek:
The error was that I filtered the issues for created vs updated and forgot to then check if the filtered issue array has actually some issues.

Am I correctly updating the bundle.json for it?


vtalas (approved):

vtalas:
@ZbynekPelunek link the issue please

Issue #265, opened 2024-11-06
[Monday - GetItem] Getting "Error fetching: Field 'title' doesn't exist on type 'ColumnValue' error


RamBachkheti:
Issue 1: [GetItem] action is showing "Error fetching: Field 'title' doesn't exist on type 'ColumnValue'" error, please see the following video:
- https://drive.google.com/file/d/1_RnEnuSZU_N7b5weL2dtFvlB24_fWGad/view?usp=sharing
<hr>

Issue 2:[AddFileToUpdate] There is no option to select the "File ID" field, please see the following video:
- https://drive.google.com/file/d/1k1cQO9szBfcpJgaFLstiqdz1F8lbREv7/view?usp=sharing

Thanks!

sayam-nasir:
1. Seems like there was a mistake in the way we were fetching the title of item as according to this article it is further layered https://community.monday.com/t/title-of-column-in-new-api/65575
I have hence fixed the query that fetches the item and it should work now

2. I have added the filepicker option in AddFileToUpdate component. It should work as well 

The PR has been raised https://github.com/clientIO/appmixer-connectors/pull/276

After the PR is approved, version 2.1.3 will have the latest fixes for Staging

RamBachkheti:
Verified the fix on staging environment.

**Test Flow:**
- https://my.stg.appmixer.com/designer/a15eab32-6bb2-46f1-a958-894a9b592b37/53fe7ccb-2916-4bc9-a59e-aa3ea24d48d8

PR #276, opened 2024-11-11
Monday.com GetItem/AddFileToUpdate (Major) Bug fixes

sayam-nasir:
fixes for https://github.com/clientIO/appmixer-connectors/issues/265

1. Seems like there was a mistake in the way we were fetching the title of item as according to this article it is further layered https://community.monday.com/t/title-of-column-in-new-api/65575
I have hence fixed the query that fetches the item and it should work now

2. I have added the filepicker option in AddFileToUpdate component. It should work as well

### DIFF
--- a/src/appmixer/monday/bundle.json
+++ b/src/appmixer/monday/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.monday",
-    "version": "2.1.2",
+    "version": "2.1.3",
     "changelog": {
         "1.0.1": [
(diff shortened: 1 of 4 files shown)


src/appmixer/monday/bundle.json
jirihofman, line 1: It is a breaking change so it should be 3.0.0.

jirihofman (changes requested):

jirihofman (approved):

PR #277, opened 2024-11-12
Google.Analytics (major) update to version 4

ZbynekPelunek:
This is an upgrade from Google Analytics 3 to 4. As found out, Google completely changed a lot of things and made a public REST API which is called `v1beta` or `v1alpha` . The Beta is considered stable with no possible breaking changes.

The upgrade will remove old components and add new ones:

- [x] Create Events (this one is not using `v1beta` route)
- [x] Generate Report


ZbynekPelunek:
When you create an event using the Create Event component, you can find the event on GA4 website here:
Screenshot 2024-11-18 115906

src/appmixer/google/analytics/CreateEvents/CreateEvents.js
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        const qeuryParams = {
+            measurement_id: measurementId, api_secret: apiSecret
+        };
+        const body = {
+            client_id: clientId, events: transformedEvents
+        };
+
+        await context.httpRequest({
```
vtalas, line 37: there has to be the `context.sendJson`, even with empty object. without it, the flow will stuck on the CreateEvent component

src/appmixer/google/analytics/CreateEvents/component.json
```diff
@@ -1,134 +1,118 @@
...
-        ]
-    },
-    "quota": {
-        "manager": "appmixer:google:analytics",
-        "maxWait": 5000,
-        "resources": "requests",
-        "scope": {
-            "userId": "{{userId}}"
-        }
+        "service": "appmixer:google:analytics"
```
vtalas, line 6: is the authentication needed? there is no authenticated call in the implementation

src/appmixer/google/analytics/CreateEvents/CreateEvents.js
```diff
@@ -0,0 +1,47 @@
...
+            throw new context.CancelError('Request can have a maximum of 25 events');
+        }
+
+        const errors = validateEvents(transformedEvents);
+
+        if (errors.length > 0) {
+            throw new context.CancelError(errors);
+        }
+
+        const qeuryParams = {
```
vtalas, line 30: typo

vtalas (changes requested):

vtalas (approved):

PR #279, opened 2024-11-14
utils.csv.AddRows (patch) optimize memory usage

jirihofman:
https://github.com/clientIO/appmixer-components/issues/1980#issuecomment-2473553187
- [x] don't load the CSV into memory
- [x] pipe it all together and pass it to engine
- [x] proper error handling

https://github.com/clientIO/appmixer-components/issues/1980#issuecomment-2473559343
- [x] call `destroy()` every time

Other stuff
- ~~change the lock duration from 60min (can't be achieved due to Appmixer 23min limit) to **5min**. That should be enough even for large files. 100MB file with 1 row added takes 15-30sec.~~ Didn't do it - keeping it backwards compatible.
- [x] Handle missing newline character in the original file

### Comparison
Below is test on local machine using 100MB csv file (100000 rows, each row 1kB). About 6sec faster. The memory increase is never grater than 10MB, even when tested on 1GB file. The same time performance increase can be observed in Appmixer with the same file.
image


src/appmixer/utils/csv/CSVProcessor.js
```diff
@@ -401,33 +400,37 @@ module.exports = class CSVProcessor {
...
-            this.writeRows(writeStream, rowsToAdd);
+            await this.loadHeaders();
+            readStream = await this.context.getFileReadStream(this.fileId);
+            writeStream = new PassThrough();
 
-            writeStream.end();
+            const promise = new Promise((resolve, reject) => {
+                const stream = pipeline(
+                    // (readStream + rowsToAdd) -> writeStream
+                    readStream.on('end', (foo) => {
```
martin-krcmar, line 417: Does this even work? How do you pipe the `readStream` to the `writeStream`? I cannot see it.
Shouldn't it look something like:
ChatGPT

martin-krcmar (changes requested):

src/appmixer/utils/csv/CSVProcessor.js
jirihofman: It worked fine. But I moved the `on` part above for clarity. Also removed the unused parameter.

vtalas (approved):

PR #286, opened 2024-11-19
ActiveCampaign (patch) multiple components fixes

ZbynekPelunek:
The UpdateDeal component requires for Deal amount to be filled, however as this is update component, it shouldn't be required on our side. My current workaround is the GET deal/dealId for the original amount and then either use it or send the new value.

### DIFF
--- a/src/appmixer/activecampaign/bundle.json
+++ b/src/appmixer/activecampaign/bundle.json
@@ -1,7 +1,15 @@
 {
     "name": "appmixer.activecampaign",
-    "version": "1.0.1",
-    "changelog": [
-        "Initial version"
-    ]
+    "version": "1.0.2",
+    "changelog": {
+        "1.0.0": [
+            "Initial version"
+        ],
+        "1.0.2": [
+            "CreateDeal: Deal amount and Currency are now required",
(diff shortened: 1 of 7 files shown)


src/appmixer/activecampaign/bundle.json
```diff
@@ -1,7 +1,15 @@
...
+    "version": "1.0.2",
+    "changelog": {
+        "1.0.0": [
+            "Initial version"
+        ],
+        "1.0.2": [
+            "CreateDeal: Deal amount and Currency are now required",
+            "CreateTask: Contact/Deal is now required",
+            "UpdateDeal: should no longer require value (Deal amount) to run",
+            "Small fixes to UpdateContact"
```
jirihofman, line 12:
```suggestion
            "CreateDeal: Deal amount and Currency are now required.",
            "CreateTask: Contact/Deal is now required.",
            "UpdateDeal: should no longer require value (Deal amount) to run.",
            "Small fixes to UpdateContact."
```

jirihofman (changes requested):

jirihofman (approved):

PR #295, opened 2024-11-28
Microsoft.SharePoint (minor) moving from onedrive API in some components

ZbynekPelunek:
There are reported issues in 4 components:
 - [x] **ListSites**
 - [x] **GetFile**
 - [x] **DownloadFile**
 - [x] **MoveFileOrFolder**

In multiple of them there was reported error "Circular reference" I found out that it is from the _onedrive-api_ library, so I had to change the components to using REST API, affected components:
**ListSites**
**GetFile**
**MoveFileOrFolder** (currenly returns an error when the move is successful)

**DownloadFile** and **GetFile** are reported to trigger multiple times for 1 file, I couldn't replicate this issue

### DIFF
--- a/src/appmixer/microsoft/sharepoint/GetFile/GetFile.js
+++ b/src/appmixer/microsoft/sharepoint/GetFile/GetFile.js
@@ -1,6 +1,4 @@
 'use strict';
-const oneDriveAPI = require('onedrive-api');
-const commons = require('../../microsoft-commons');
 
 module.exports = {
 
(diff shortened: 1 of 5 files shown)


ZbynekPelunek:
The **MoveFileOrFolder** currenly returns an error when the move is successful. I have opened a question on Microsoft knowledge base https://learn.microsoft.com/en-us/answers/questions/2125407/move-a-driveitem-to-a-new-folder-api-returns-404-o

ZbynekPelunek:
This is the final answer from Microsoft:

> After analysis, because this movement is already a cross-drives movement, but according to the document description, this behavior is not allowed:
> 
> image
> 
> So, **an error is expected**, but the fact that the file can be moved successfully indicates that the endpoint is not perfect, and you can either give feedback [here ](https://feedbackportal.microsoft.com/feedback/forum/ebe2edae-97d1-ec11-a7b5-0022481f3c80)or, if you don't mind, continue to use the endpoint until it is fixed.
> 
> But for moving files across Drives, we recommend that you use the [Copy item](https://learn.microsoft.com/en-us/graph/api/driveitem-copy?view=graph-rest-1.0&tabs=http) endpoint.

So technically the error is correct but the result is wrong. It either should not return the error while moving the file across drives or return an error and don't move the item. I don't think that we should check if someone is moving a file from one drive to another, so I will just mention it in the component that when you do that, an error is expected.

jirihofman (approved):

PR #299, opened 2024-12-04
Webflow (major) Update of API from v1 to v2

sayam-nasir:


sayam-nasir:
@vtalas I have removed the bluebird dependency and also improved the authorization

vtalas (approved):

PR #302, opened 2024-12-09
gmail (major) Additional fixes and improvements

DavidDurman:


src/appmixer/google/gmail/DownloadAttachment/component.json
```diff
@@ -32,40 +31,27 @@
                         "type": "string"
                     }
                 },
-                "required": ["messageId", "attachmentId", "fileName"]
+                "required": ["messageId", "attachmentId"]
```
vtalas, line 34: `fileName` is not reuquired, however flow throws the `missing file name param` if it's not provided 
image

vtalas (changes requested):
also consider renaming gmail-commons.js to lib.js

vtalas:
finalized in https://github.com/clientIO/appmixer-connectors/pull/373

PR #312, opened 2024-12-16
OpenAI (patch) CreateCompletion, RetrieveModel

ZbynekPelunek:
I have added dynamic list of models to select in **CreateCompletion** component, however some of the models return an error that they are not supported for this endpoint.


ZbynekPelunek:
Just to clarify - the CreateCompletion component has a lot of parameters, I have tested some combination with various values and found out that "stream" was causing the response to be stringified JSON, even the description of this parameter is: 

> Whether to stream back partial progress. If set, tokens will be sent as data-only [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format) as they become available, with the stream terminated by a data: [DONE] message

so I have removed it and kept the rest.

src/appmixer/openai/core/createCompletion/createCompletion.js
```diff
@@ -1,57 +1,26 @@
...
     receive: async function(context) {
+        const { content } = context.messages.in;
 
-        const { data } = await this.httpRequest(context);
-
-        return context.sendJson(data, 'out');
-    },
-
-    httpRequest: async function(context) {
+        context.log({ step: 'auth', auth: context.auth });
```
vtalas, line 8: remove this, we mustn't log auth info into logs

vtalas (changes requested):

src/appmixer/openai/core/createCompletion/createCompletion.js
ZbynekPelunek: of course, just a leftover from my development, thanks for catching that

vtalas (approved):

PR #313, opened 2024-12-16
ActiveCampaign.CreateTask (patch)

ZbynekPelunek:
The **CreateTask** component required both _contactId_ and _dealId_, however it should be either or.

### DIFF
--- a/src/appmixer/activecampaign/bundle.json
+++ b/src/appmixer/activecampaign/bundle.json
@@ -1,7 +1,15 @@
 {
     "name": "appmixer.activecampaign",
-    "version": "1.0.1",
-    "changelog": [
-        "Initial version"
-    ]
+    "version": "1.0.2",
+    "changelog": {
+        "1.0.0": [
+            "Initial version"
+        ],
+        "1.0.2": [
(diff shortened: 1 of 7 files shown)


src/appmixer/activecampaign/bundle.json
```diff
@@ -1,7 +1,15 @@
...
-    "changelog": [
-        "Initial version"
-    ]
+    "version": "1.0.2",
+    "changelog": {
+        "1.0.0": [
+            "Initial version"
+        ],
+        "1.0.2": [
+            "CreateDeal: Deal amount and Currency are now required.",
```
vtalas, line 9: making fields required is a breaking change (it should be v2.0.0)

ZbynekPelunek: I understand but what if the component didnt work without it before? I mean you couldnt create a deal without Deal amount and Currency

jirihofman (approved):

vtalas (approved):

vtalas:
fixes https://github.com/clientIO/appmixer-components/issues/1920#issuecomment-2543935315

PR #314, opened 2024-12-16
AzureDocumentIntelligence.ClassifyDocument - better tooltips

ZbynekPelunek:


vtalas (changes requested):
@ZbynekPelunek  please update the bundle.json as well

ZbynekPelunek:
> @ZbynekPelunek please update the bundle.json as well

Alright, I wasn't sure if I should do it as it is only tooltip change, thanks for the confirmation.

src/appmixer/azuredocumentintelligence/classifiers/ClassifyDocumentFromStream/ClassifyDocumentFromStream.js
```diff
@@ -0,0 +1,38 @@
...
+        const { endpoint, apiKey } = context.config;
+        const { classifierId, fileId } = context.messages.in.content;
+
+        const client = DocumentIntelligence(endpoint, { key: apiKey });
+
+        let options;
+
+        const fileInfo = await context.getFileInfo(fileId);
+        const fileStream = await context.getFileReadStream(fileId);
+        options = {
```
jirihofman, line 19: Can `options` be const?

```suggestion
        const fileInfo = await context.getFileInfo(fileId);
        const fileStream = await context.getFileReadStream(fileId);
        const options = {
```

jirihofman (approved):

vtalas (approved):

PR #335, opened 2025-01-09
Utils.Files (minor) ArchiveFile: new component

ZbynekPelunek:
Seems to be working using the **context.archiveFile()** function on QA.
Things to check:

- [x] Ouput
- [x] quota

Currently, there is no indication on the UI if the file is archived, however, it is possible to be checked in the browser Network tab, while loading the Files page.

### DIFF
--- a/src/appmixer/utils/files/ArchiveFile/component.json
+++ b/src/appmixer/utils/files/ArchiveFile/component.json
@@ -1,0 +1,49 @@
+{
+    "name": "appmixer.utils.files.ArchiveFile",
+    "author": "Zbynek Pelunek <zbynek.pelunek@client.io>",
+    "description": "Archive file from File ID.",
+    "version": "1.0.0",
+    "private": false,
+    "quota": {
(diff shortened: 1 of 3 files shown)


src/appmixer/utils/files/ArchiveFile/component.json
```diff
@@ -0,0 +1,49 @@
...
+                    }
+                },
+                "required": [
+                    "fileId"
+                ]
+            },
+            "inspector": {
+                "inputs": {
+                    "fileId": {
+                        "type": "text",
```
vtalas, line 29: @ZbynekPelunek  can't we use the `"type": "filepicker",` ?

ZbynekPelunek:
@vtalas just to be sure, is the "quota" used there correctly? I got it from Download File, but I am not sure if it actually does the same thing. If it only flags already stored file as archived I assume the quota is not needed.

vtalas (approved):

PR #343, opened 2025-01-16
google.drive (minor) FindFilesOrFolders now support recursive search …

DavidDurman:
…and orderBy parameter


vtalas (changes requested):
FindFilesOrFolders 
- query: "fasada-doklad", recursive: true, all other inputs intact 
image
 
sems the `Folder` input is required. works fine when a Folder is selected

PR #347, opened 2025-01-17
utils.converters.PDF2Text (new) convert PDF to text

DavidDurman:


src/appmixer/utils/converters/converters.js
```diff
@@ -95,5 +97,26 @@ module.exports = {
...
+        if (!pdfjslib) {
+            pdfjslib = await import('pdfjs-dist/legacy/build/pdf.mjs');
+        }
+        const readStream = await context.getFileReadStream(sourceFileId);
+        let text = '';
+        // Unfortunately, `data` can only be a string, buffer or TypedArray.
+        // getDocument() does not work over streams.
+        const loadingTask = pdfjslib.getDocument({ data: await arrayBuffer(readStream) });
+        const pdfDoc = await loadingTask.promise;
+        for (let i = 0; i < pdfDoc.numPages; i++) {
```
vtalas, line 113: based on this code, it seems that the entire PDF file is not loaded into memory all at once; instead, it is processed page by page. Is that correct? 

However, the entire output text is stored in memory. I believe we should consider sending the output text chunks into a stream, or we should introduce a strict limit on the input size of the PDF file. With larger PDF files containing a lot of text, there's a possibility of reaching the memory limit on the instance.

vtalas (changes requested):

PR #350, opened 2025-01-17
GoogleCalendar.CreateEvent (patch) Fix to show only write access calendars

sayam-nasir:


src/appmixer/google/calendar/CreateEvent/component.json
```diff
@@ -5,8 +5,7 @@
     "auth": {
         "service": "appmixer:google",
         "scope": [
-            "https://www.googleapis.com/auth/calendar",
-            "https://www.googleapis.com/auth/calendar.readonly"
+            "https://www.googleapis.com/auth/calendar.events"
```
vtalas, line 8: `https://www.googleapis.com/auth/calendar.events` is restricted scope, requires verification 
image

vtalas (changes requested):
@sayam-nasir  it's bit mor complicated with the scopes, it's not properly set. I tried to summarize potential issue here https://github.com/clientIO/appmixer-components/issues/2085

in our latest Google app, we have the `calendar.readonly` and  `calendar` approved.

vtalas (approved):

PR #372, opened 2025-02-11
Kafka (patch) add more SSL options

jirihofman:
Part of https://github.com/clientIO/appmixer-components/issues/2093

Options to connect with
- [x] SASL + `rejectUnathorized`
- 3 certificates, no SASL
    - [x] triggered via webhook after `start` finished
    - [x] called immediately on flow start

### New SSL options added
- rejectUnathorized
- 3x cert files

image


src/appmixer/kafka/bundle.json
```diff
@@ -1,9 +1,13 @@
 {
     "name": "appmixer.kafka",
-    "version": "1.0.1",
+    "version": "1.0.2",
```
vtalas, line 3: it should be 1.1.0, I suppose, as there is new feature

vtalas (changes requested):

vtalas (approved):

PR #388, opened 2025-02-24
Slack (hotfix) 2.4.1 (into 2.4.2)

jirihofman:
Fix for https://github.com/clientIO/appmixer-components/issues/2139

### Before
```js
users = await client.listUsers();
```

### After
```js
users = await client.listUsers({ limit: 999 });
```


jirihofman:
Done in #386 and manually published

PR #510, opened 2025-05-13
Utils (patch) POST/PUT added output processing

sayam-nasir:
Added output processing in case api doesn't return a json response

## Summary by CodeRabbit

- **New Features**
	- Enhanced HTTP response handling for POST and PUT requests, providing more detailed and structured information in responses, including status code, request details, headers, and parsed content-type.
- **Improvements**
	- Responses for binary and form-data requests are now consistently formatted for easier interpretation.
	- Improved robustness when handling API responses that do not return valid JSON in binary data scenarios.

### DIFF
--- a/src/appmixer/utils/http/Post/Post.js
+++ b/src/appmixer/utils/http/Post/Post.js
@@ -1,3 +1,4 @@
+const contentTypeUtil = require('content-type');
 const request = require('../http-commons');
 const FormData = require('form-data');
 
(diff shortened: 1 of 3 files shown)


sayam-nasir:
@vtalas The reason you see repeated code in both files is because I didn't want to make drastic changes to the http-commons as it is coded to export 'request' which is being used in 5 components

vtalas (approved):

PR #553, opened 2025-06-05
Everart (new)

vtalas:
- [x] Create Model
- [x] Create Model Generations
- [x] Find Models
- [x] Get Generation
- [x] Get Model 
- [x] Update Model 

## Summary by CodeRabbit

- **New Features**
  - Introduced EverArt integration, enabling users to create, update, and manage models, as well as generate images from text or image prompts.
  - Added components for finding models, retrieving a model or generation by ID, and updating model details.
  - Implemented support for batch image generation with flexible output options and webhook notifications.
  - Included authentication setup for secure API key management.
- **Documentation**
  - Provided detailed usage instructions and example commands for all new EverArt features and components.


ZbynekPelunek:
Function-wise I believe it is fine, even though it's harder to test when we are out of credits on the account. However, I would unify the labels of IDs, for example in GenerateImage, you have "Model Id", then in UpdateModel, you have just "Id". It would be better if it was just "Model ID" or "Foo ID" everywhere.

vtalas:
> Function-wise I believe it is fine, even though it's harder to test when we are out of credits on the account. However, I would unify the labels of IDs, for example in GenerateImage, you have "Model Id", then in UpdateModel, you have just "Id". It would be better if it was just "Model ID" or "Foo ID" everywhere.

@ZbynekPelunek good point, updated here https://github.com/clientIO/appmixer-connectors/pull/553/commits/13bf3d68803f1d91f2501fa193996b0ac3b18955

src/appmixer/everart/core/GetGeneration/GetGeneration.js
```diff
@@ -0,0 +1,19 @@
...
+        const { data } = await context.httpRequest({
+            method: 'GET',
+            url: `https://api.everart.ai/v1/generations/${id}`,
+            headers: {
+                'Authorization': `Bearer ${context.auth.apiKey}`
+            }
+        });
+
+
+        console.log(data);
```
ZbynekPelunek: Sorry, one last thing I have just noticed, I assume the console.log() in this case is not intended. Once removed it will be ready for merge.

src/appmixer/everart/core/FindModels/FindModels.js
```diff
@@ -0,0 +1,45 @@
...
+    'estimated_completed_at': { 'type': 'string', 'title': 'Estimated Completed At' },
+    'thumbnail_url': { 'type': 'string', 'title': 'Thumbnail Url' }
+};
+
+module.exports = {
+    async receive(context) {
+        const { search, status, outputType } = context.messages.in.content;
+
+        if (context.properties.generateOutputPortOptions) {
+            return lib.getOutputPortOptions(context, outputType, schema, { label: 'Models', value: 'models' });
```
ZbynekPelunek: The property `value` here shouldn't be 'models' but 'result' so it matches with the lib.generated.js

src/appmixer/everart/core/FindModels/FindModels.js
```diff
@@ -0,0 +1,45 @@
...
+                'Authorization': `Bearer ${context.auth.apiKey}`
+            },
+            params: queryParams
+        });
+
+        if (!data.models.length) {
+            return context.sendJson({}, 'notFound');
+        }
+
+        return lib.sendArrayOutput({ context, records: data.models, outputType, arrayPropertyValue: 'models' });
```
ZbynekPelunek: `arrayPropertyValue` can be removed as it is no longer part of sendArrayOutput()

ZbynekPelunek (approved):

PR #646, opened 2025-07-16
GoogleTasks (patch) Silencing 400 errors if TaskList not present

sayam-nasir:
For fetching a list of tasks, for dynamic input, we are required to have a List ID present. If the list ID is selected from a tag in a flow and not the list dropdown itself, then there is an error on the task endpoint. I have silenced the 400 something error in this case so we can run the flow with input tags

## Summary by CodeRabbit

* **New Features**
  * The "status" field when updating a task is now a dropdown with predefined options ("Needs Action" and "Completed") for easier selection.

* **Bug Fixes**
  * Improved error handling when listing tasks: client-side errors now return an empty list instead of causing failures.

* **Documentation**
  * Tooltips for task selection fields have been updated to clarify that the task must already exist in the specified task list.

* **Chores**
  * Updated the package version number and changelog entry.

### DIFF
--- a/src/appmixer/googleTasks/bundle.json
+++ b/src/appmixer/googleTasks/bundle.json
@@ -1,8 +1,8 @@
 {
     "name": "appmixer.googleTasks",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "changelog": {
-        "1.0.0": [
+        "1.0.1": [
             "Initial version"
         ]
     }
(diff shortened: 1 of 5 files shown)


vtalas (changes requested):
`tasklist` should be required in the input schema, the call obviously doesn't make sense without it, and there should be

```
if (!tasklist) {
    throw new context.CancelError ... 
}
```
We realize this check is necessary because if the `tasklist` is passed into the component as a variable, we want to avoid making a request that we know will fail.

FYI @ZbynekPelunek  @jirihofman

vtalas (approved):

PR #666, opened 2025-07-24
Wiz.UploadScan(minor): improvements 

vtalas:
## Summary by CodeRabbit

* **New Features**
  * Improved document upload batching and scheduling using unique IDs for each document.
  * Enhanced logging for upload scheduling and lock handling.
  * Added local state management simulation in test utilities for more isolated testing.

* **Bug Fixes**
  * More robust handling of document uploads with improved state management and locking mechanisms.

* **Tests**
  * Updated tests to properly await asynchronous scheduling calls, ensuring accurate timing assertions.

* **Chores**
  * Updated dependency list with a new UUID package.
  * Incremented component and bundle version numbers reflecting stability and performance improvements.


src/appmixer/wiz/core/UploadScan/UploadScan.js
```diff
@@ -30,41 +30,125 @@ module.exports = {
 
         const { threshold, scheduleValue } = context.properties;
 
-        let lock;
-        try {
+        if (context.messages.timeout) {
+            await this.scheduleDrain(context);
```
martin-krcmar, line 34: Consider this scenario:
- timeout hits
- another drain is scheduled
- system crashes/restarts right after the drain is scheduled
- You have 2 timeouts scheduled; they both could enter the `receive()` in the future, processing the same documents.

martin-krcmar: I can see that there is a lock in the `prepareForSend`, but if 2 processes get here. The first one would lock and prepare, then release the lock, and the second one will get the lock and do the same. They both could send the same documents.

vtalas: > I can see that there is a lock in the `prepareForSend`, but if 2 processes get here. The first one would lock and prepare, then release the lock, and the second one will get the lock and do the same. They both could send the same documents.

I shouldn't happen, as `prepareForSend` moves data ([source](https://github.com/clientIO/appmixer-connectors/pull/666/files/8da261248760f41905a5b17651e1f606a3fd3c1e#diff-31f544850b8b056fd5d6dce229d22c296ef1f79ef3af79aefd2c4b3e15b57625R103)) from `documents` to `documents-upload-batch`. When the 2nd process arrives, the `documents-upload-batch` is already define and it skips the preparation https://github.com/clientIO/appmixer-connectors/pull/666/files/8da261248760f41905a5b17651e1f606a3fd3c1e#diff-31f544850b8b056fd5d6dce229d22c296ef1f79ef3af79aefd2c4b3e15b57625R69

vtalas: > Consider this scenario:
> 
> * timeout hits
> * another drain is scheduled
> * system crashes/restarts right after the drain is scheduled
> * You have 2 timeouts scheduled; they both could enter the `receive()` in the future, processing the same documents.

Yes, that’s correct. However, it shouldn’t process the same documents again—there will just be an additional call. The logic in prepareForSend should prevent duplicate processing. 
There should be the same logic with the `timeoutId` that's already in the Scheduler, component, right?

implemented here https://github.com/clientIO/appmixer-connectors/pull/666/commits/a3134ead6ace99d5a97d3ac322a8ace3e82d810b

martin-krcmar: Well. You could add that, but I wouldn't complicate it. If it is secured, that the same documents cannot be processed (sent) twice, then it's fine. Although if that happens (and it can) - that the component/system crashes in the middle of the `receive()`, you will end up with 2 timeouts for the same component - forever - until you stop the flow. So detecting that could save you some troubles.

PR #700, opened 2025-08-19
Groq CreateTranscription/CreateTranslation. Removal of responseFormats as we only have JSON

sayam-nasir:
Groq's API for Create Transcription and Create Translation supports 3 response formats
1. Text
         In this, the response is just a plain text 
         
image

2. JSON
      In this, the response is JSON
      
image

3. Verbose JSON
      In this, the output is JSON but has alot more output fields
      
image

We cannot have 2 different output messages in one component, and also cannot have plain text like output so I am keeping JSON only

## Summary by CodeRabbit

- Refactor
  - Migrated Groq components to the appmixer.ai.groq namespace and updated component identifiers, auth, quotas, and internal API paths.
- Bug Fixes
  - Added required-input validation for transcription, translation, prompt sending, and API calls.
  - Removed "response format" input from Transcription and Translation flows.
- Style
  - Improved Temperature field label and tooltip.
- Chores
  - Updated and reorganized package/module manifests; removed deprecated Groq package files.


vtalas (dismissed):
@sayam-nasir this request is not related to this PR, however it should be done anyway
CreateChatCompletion -> SendPrompt
whole groq connector should be moved to the `ai` group

src/appmixer/ai/groq/service.json
```diff
@@ -1,5 +1,5 @@
 {
```
ZbynekPelunek: I tried to publish the component to QA but ran into an error: 

> [ERROR]:  Publishing failed with following error:
> appmixer/ai/service.json name property doesn't match the file path

Since this is under AI now, I think it should have `module.json` file instead of `service.json`

ZbynekPelunek (changes requested):
Everything seems fine, just replace `service.json` with `module.json`

ZbynekPelunek (commented):
Have you tried to publish it? I am getting an error:

> [ERROR]:  Publishing failed with following error:
> Error in component core
> Cannot find module '/tmp/tmp-7-wMO52OVkgpQq/appmixer/ai/groq/core/core'

sayam-nasir:
@ZbynekPelunek the components were still under core, which is now fixed and it is being published successfully

vtalas (changes requested):
remove the appmixer/src/groq

sayam-nasir:
> remove the appmixer/src/groq

@vtalas I believe it's already removed, because in this PR I cannot find it anywhere in the files, I did cut and moved the folder under ai folder before updating it.

In the changed files, it is showing 2 files removed from the groq folder

ZbynekPelunek (approved):
I can also confirm that I don't see the groq under just `./appmixer/` subfolder, only under `./appmixer/ai/`
And can publish the connector successfully now

vtalas (approved):

PR #706, opened 2025-08-20
LemonSqueezy - test flows

ZbynekPelunek:
Added 2 test flows: 
1. focused on customers (create + update) with asserts.
2. focused on Subscriptions (with assert), Orders and Products


vtalas (approved):

vtalas (changes requested):
remove the zip bundle

vtalas (approved):

PR #763, opened 2025-09-09
Utils.SaveToFile (patch) Fixed Error

sayam-nasir:
## Summary by CodeRabbit

* **Bug Fixes**
  * CSV export for non-flattened key/value data now includes a header, correct quoting per-value, and stable generation.
  * Mitigated CSV/formula injection by prefixing dangerous values.
  * Resolved CSV-processing errors that caused failures for certain record formats.

* **Chores**
  * Switched to the maintained CSV package and bumped the storage bundle version to 2.0.1 with changelog.


vtalas (changes requested):
@sayam-nasir it freezes on most of the storages, exceptions are likely not properly propagated. 
for example 

Image

sayam-nasir:
@vtalas there was a bug in the case when filetype is csv and flatten value toggle is true

vtalas (approved):

PR #778, opened 2025-09-22
Big Commerce. (minor) Updates 

sayam-nasir:
## Summary by CodeRabbit

- New Features
  - CreateCustomer now outputs Customer Id.
  - Added List Order Status, List Brands, and List Product Categories components for dynamic selections.
  - FindOrders and FindProducts support selecting status/brand from dynamic lists; product categories input now supports multiselect and normalization.
  - Input normalization for multiselects added.

- Bug Fixes
  - Date filters (customers/products) now use ISO 8601 without milliseconds.

- Chores
  - BigCommerce bundle bumped to 1.0.1.

- Tests
  - Unit tests added for multiselect normalization.

### DIFF
--- a/src/appmixer/bigCommerce/core/CreateCustomer/component.json
+++ b/src/appmixer/bigCommerce/core/CreateCustomer/component.json
@@ -125,6 +125,13 @@
             "name": "out",
             "options": [
                 {
+                    "label": "Customer Id",
+                    "value": "id",
+                    "schema": {
+                        "type": "string"
+                    }
+                },
+                {
                     "label": "Email",
                     "value": "email",
                     "schema": {
--- a/src/appmixer/bigCommerce/core/FindOrders/component.json
+++ b/src/appmixer/bigCommerce/core/FindOrders/component.json
@@ -24,7 +24,7 @@
                         "type": "number"
                     },
                     "status_id": {
-                        "type": "number"
+                        "type": "string"
                     },
                     "payment_method": {
                         "type": "string"
--- a/src/appmixer/bigCommerce/core/FindOrders/component.json
+++ b/src/appmixer/bigCommerce/core/FindOrders/component.json
@@ -61,9 +61,15 @@
                         "index": 1
                     },
                     "status_id": {
-                        "type": "number",
-                        "tooltip": "Filter orders by status ID. Examples: 0 = Incomplete, 1 = Pending, 2 = Shipped, 3 = Partially Shipped, 4 = Refunded, 5 = Cancelled, 6 = Declined, 7 = Awaiting Payment, 8 = Awaiting Pickup, 9 = Awaiting Shipment, 10 = Completed, 11 = Awaiting Fulfillment, 12 = Manual Verification Required, 13 = Disputed, 14 = Partially Refunded. See full list: https://developer.bigcommerce.com/docs/rest-management/orders/order-status#get-all-order-statuses",
-                        "label": "Status ID",
+                        "type": "text",
+                        "tooltip": "Select the order status from the list or enter the order status ID.",
+                        "label": "Order Status",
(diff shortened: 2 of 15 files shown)


src/appmixer/bigCommerce/core/FindProducts/component.json
```diff
@@ -79,9 +79,15 @@
...
                         "label": "Brand ID",
+                        "source": {
+                            "url": "/component/appmixer/bigCommerce/core/ListBrands?outPort=out",
+                             "data": {
+                                "transform": "./ListBrands#brandToSelectArray"
+                            }
+                        },
                         "index": 3
                     },
                     "categories": {
```
ZbynekPelunek, line 93: Could you also please add dynamic select for categories using this endpoint: https://developer.bigcommerce.com/docs/rest-catalog/category-trees/categories#get-all-categories
The same reason as Brand ID, the filter option right now is useless unless you somehow keep track of your category IDs which you can get only through this component. I know it's not as simple as the Brand ID because it will be a multiselect, but its either that or removing the property.

src/appmixer/bigCommerce/core/FindOrders/component.json
```diff
@@ -61,9 +61,15 @@
...
+                        "label": "Order Status",
+                        "source": {
+                            "url": "/component/appmixer/bigCommerce/core/ListOrderStatus?outPort=out",
+                            "data": {
+                                "transform": "./ListOrderStatus#statusToSelectArray"
+                            }
+                        },
                         "index": 2
                     },
                     "payment_method": {
```
ZbynekPelunek, line 75: The Order Status dynamic select works great. Just a question, why also not make `payment_method` into a select with the options mentioned in the tooltip? Could be a 'text' - type select which allows you to either select option below or write your own. Or is the payment method customizable so it can only be dynamic select?

src/appmixer/bigCommerce/core/CreateCustomer/component.json
```diff
@@ -124,6 +124,13 @@
         {
             "name": "out",
             "options": [
+                {
```
ZbynekPelunek, line 127: I can confirm that this works, I just don't understand the magic how :D I could clearly see in their documentation that there is no ID in the API response so we agreed upon calling the GET endpoint to retrieve it, but since there are no changes to CreateCustomer.js and it still has only 1 call, how is it possible that it returns ID now? Just curious

ZbynekPelunek (dismissed):

src/appmixer/bigCommerce/core/FindOrders/component.json
sayam-nasir: [Payment methods](https://developer.bigcommerce.com/docs/rest-payments/methods) can be gotten but we need order ID for that, so not possible in this inspector as this is FindOrders itself.
Payment methods direct API is [deprecated](https://developer.bigcommerce.com/docs/rest-payments/methods-deprecated)
We cannot create a "type: "text"  in the inspector and have a dropdown list, either the "type" has to be a select or it needs to have a "source" if we want to have a dropdown list.
Creating a whole component to list 4 hardcoded payment methods doesn't make sense compared to providing a tooltip. Also, the payment methods input doesn't require ID, just a display name of payment method so I believe this one can be taken care by tooltip text.

src/appmixer/bigCommerce/core/CreateCustomer/component.json
sayam-nasir: I also didn't add it before because the docs do not mention it, but while I was adding the new endpoint in the behaviour file, I discovered that the original endpoint does return an ID. I just mapped it and it works

vtalas (approved):

vtalas (approved):

PR #780, opened 2025-09-23
HubSpot (fix) - ensure webhook configuration outside of AuthHub

jirihofman:
Fix for https://github.com/clientIO/appmixer-components/issues/2513
- please look at the alternative solution as well

This might be just adding more complexity when none is needed.

## Summary by CodeRabbit

- Bug Fixes
  - Fixed trigger reliability with AuthHub.
  - Added startup webhook verification that fails fast when HubSpot webhook target mismatches.
  - Ensures webhook is registered when missing (404) and surfaces clear errors on other HubSpot failures.

- Chores
  - Bumped HubSpot bundle version to 4.2.2 and added changelog entry.

- Tests
  - Expanded tests for webhook verification, mismatch handling, and 404-based registration behavior.

### DIFF
--- a/src/appmixer/hubspot/BaseSubscriptionComponent.js
+++ b/src/appmixer/hubspot/BaseSubscriptionComponent.js
@@ -23,6 +23,75 @@ class BaseSubscriptionComponent {
         throw new Error('Must be extended to return subscriptions');
     }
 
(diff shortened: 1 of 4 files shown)


vtalas (approved):

vtalas:
verified on STG

PR #787, opened 2025-09-26
Square (patch) Fixes

sayam-nasir:
## Summary by CodeRabbit

- New Features
  - Added List Customer Groups component to fetch customer groups.
  - Find Customers: added notFound output when no records are returned.

- Refactor
  - Find Customers inputs accept string or array for creationSource/groupIds; improved filter handling and non-error empty-result behavior.

- Documentation
  - Expanded phone number guidance in Create/Update Customer tooltips.
  - Minor label/description wording tweaks (e.g., Customer Id).

- Chores
  - Version bumped to 1.0.1.

### DIFF
--- a/src/appmixer/square/core/FindCustomers/component.json
+++ b/src/appmixer/square/core/FindCustomers/component.json
@@ -36,16 +36,10 @@
                         "type": "string"
                     },
                     "creationSource": {
-                        "type": "array",
-                        "items": {
-                            "type": "string"
-                        }
+                        "type": "string"
                     },
                     "groupIds": {
-                        "type": "array",
-                        "items": {
-                            "type": "string"
-                        }
+                        "type": "string"
                     },
                     "sortField": {
                         "type": "string"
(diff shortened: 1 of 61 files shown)


src/appmixer/square/core/FindCustomers/component.json
```diff
@@ -108,19 +102,25 @@
...
-                            { "label": "Square Dashboard", "value": "DASHBOARD" },
-                            { "label": "Square Point of Sale", "value": "SQUARE_POS" },
-                            { "label": "Terminals API", "value": "TERMINALS_API" },
-                            { "label": "Terminal", "value": "TERMINAL" },
-                            { "label": "Billing", "value": "BILLING" },
-                            { "label": "Bookings", "value": "BOOKINGS" },
-                            { "label": "Marketing", "value": "MARKETING" },
+                            { "label": "Other", "value": "OTHER" },
+                            { "label": "Appointments", "value": "APPOINTMENTS" },
+                            { "label": "Coupon", "value": "COUPON" },
```
ZbynekPelunek:
```suggestion
                            { "clearItem": true, "label": "<empty>"},
                            { "label": "Other", "value": "OTHER" },
                            { "label": "Appointments", "value": "APPOINTMENTS" },
                            { "label": "Coupon", "value": "COUPON" },
```
It should also have an option to unselect the value using the `"clearItem": true` as the selection is not required and I might want to not filter it by any Creation Source

src/appmixer/square/core/FindCustomers/component.json
```diff
@@ -108,19 +102,25 @@
...
+                            { "label": "Merge", "value": "MERGE" },
+                            { "label": "Online Store", "value": "ONLINE_STORE" },
+                            { "label": "Instant Profile", "value": "INSTANT_PROFILE" },
+                            { "label": "Terminal", "value": "TERMINAL" },
+                            { "label": "Third Party", "value": "THIRD_PARTY" },
+                            { "label": "Third Party Import", "value": "THIRD_PARTY_IMPORT" },
+                            { "label": "Unmerge Recovery", "value": "UNMERGE_RECOVERY" }
                         ]
                     },
                     "groupIds": {
```
ZbynekPelunek, line 187: I see that you added ListCustomerGroups component which is great just is there a reason why is it not used here for dynamic select?

src/appmixer/square/core/GetCustomer/component.json
```diff
@@ -42,7 +42,7 @@
             "name": "out",
             "options": [
                 {
-                    "label": "Customer Id",
+                    "label": "Id",
```
ZbynekPelunek:
```suggestion
                    "label": "Customer Id",
```
It was fine before

ZbynekPelunek (changes requested):

src/appmixer/square/core/FindCustomers/component.json
sayam-nasir: It can be used here, you are right but the reason for this was
1. We prefer having a "Find x" component that can be used as a source in the flow over a dynamic input field.
2. In this case there can't be a Find Groups hence a List Groups in it's place.

Nevertheless, I can add that as a source endpoint here for a better user experience.

Thanks!

src/appmixer/square/core/FindCustomers/component.json
sayam-nasir: That's a great point, Let me try and add for all non required select fields if possible.

Thanks!

ZbynekPelunek (approved):

PR #788, opened 2025-09-29
HELPSCOUT (NEW)

sayam-nasir:


vtalas (changes requested):
@sayam-nasir  please move the unit tests to the artifacts folder. Check for example the https://github.com/Appmixer-ai/appmixer-connectors/tree/dev/src/appmixer/freshdesk/artifacts/test for reference

vtalas (approved):

PR #789, opened 2025-09-29
Front (new)

ZbynekPelunek:


ZbynekPelunek:
@coderabbitai review

src/appmixer/front/lib.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+};
+
+/**
+ * @param {array} array
+ * @returns {string}
+ */
+const toCsv = (array) => {
+
+    const headers = Object.keys(array[0]);
```
jirihofman, line 142: Add this fix from before: https://github.com/Appmixer-ai/appmixer-cli/pull/285/files#diff-4512df4d6ae77c395c6a99b3e9ab7a45ba7a0b5587e68fd76a4532e24f8ce504R110-R113

src/appmixer/front/service.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "appmixer.front",
+    "label": "Front",
+    "category": "applications",
+    "description": "Front is a customer communication platform. This connector uses Front Core API (https://api2.frontapp.com) to manage contacts, conversations, and messages.",
```
jirihofman: Is this up-to-date? I can see other entities, eg Teams, Tags, ...

jirihofman (changes requested):
I noticed few `context.log()` calls of HTTP requests (not sure where due too many files 😆 ) . We should either have them everywhere or not at all. IMHO remove them.

ZbynekPelunek:
> I noticed few `context.log()` calls of HTTP requests (not sure where due too many files 😆 ) . We should either have them everywhere or not at all. IMHO remove them.

Those were just some leftovers from my testing, thanks for catching that. I log the response like this so I can then use the data to generate the most up-to-date output port options.

jirihofman (approved):

PR #807, opened 2025-10-08
Slack (patch) - Tasks + AuthHub fixes

jirihofman:
- [x] do not validate payload on tenant when it is forwarded from AuthHub. it is already validated there

## Summary by CodeRabbit

- Bug Fixes
  - Correctly handle Slack interaction requests forwarded via AuthHub, ensuring they’re processed without unnecessary signature checks while retaining validation for direct Slack requests. Improves reliability of tenant interactions.

- Tests
  - Added coverage to verify behavior for AuthHub-forwarded Slack interactions.

- Chores
  - Bumped Slack bundle version to 5.1.2 and updated the changelog accordingly.

### DIFF
--- a/src/appmixer/slack/bundle.json
+++ b/src/appmixer/slack/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.slack",
-    "version": "5.1.1",
+    "version": "5.1.2",
     "engine": ">=6.0.0",
     "changelog": {
         "1.0.1": [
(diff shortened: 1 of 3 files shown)


jirihofman:
Waiting on https://github.com/clientIO/appmixer-core/issues/3164 to be able to work with Slack connector properly in multi-pod environment. After that I noticed a similar issue. Accepting/Rejecting Slack task sometimes fails with 401 (in Newrelic logs). After several tries (when particular pod is hit?) it succeeds.

vtalas (approved):

PR #809, opened 2025-10-09
Slack Tasks (patch) - move task files to root folder

jirihofman:
Based on https://github.com/Appmixer-ai/appmixer-core/issues/3164#issuecomment-3384753976

This fixes errors on multi-pod tenants


jirihofman:
Verified on multi-pod STG with AuthHub. Works as expected.

vtalas (approved):

PR #839, opened 2025-10-22
GoogleDrive (patch) - better File ID validation + label

### LINKED ISSUE #808: google.drive.uploadFile: file not found error
https://appmixer.freshdesk.com/a/tickets/9363

Customer reported an issue while uploading a file to gdrive: 

Image

root cause is described in the freshdesk ticket

we should be able handle this error, possible solution (by Claude)
```
if (files.length > 0) {
    const file = files[0];
    try {
        response = await drive.files.update({
            fileId: file.id,
            quotaUser: userId,
            media: {
                mimeType: contentType,
                body: fileStream
            },
            fields: 'id, name, mimeType, webViewLink, createdTime'
        });
    } catch (error) {
        // File exists in search but not accessible - treat as new file
        if (error.code === 404 || error.status === 404) {
            await context.log({ 
                step: 'file-not-found-fallback', 
                fileId: file.id,
                message: 'File found in search but not accessible, creating new file'
            });
            response = await drive.files.create({
                quotaUser: userId,
                resource,
                media: {
                    mimeType: contentType,
                    body: fileStream
                },
                fields: 'id, name, mimeType, webViewLink, createdTime'
            });
        } else {
            throw error;
        }
    }
}
```

UPDATE:
the customer reported that he selected a folder from the Shared Drive. When he selects My Drive, then it works even with replace existing is enabled. Thus, the bug is in the Shared Drive folder selection.

ZbynekPelunek:
When the File ID is incorrect, it now rather cancel the flow with a descriptive error.
Also updated the UploadFile - File ID field to "`Appmixer File ID`"

### DIFF
--- a/src/appmixer/google/drive/UploadFile/component.json
+++ b/src/appmixer/google/drive/UploadFile/component.json
@@ -49,8 +49,8 @@
                 "fileId": {
                     "type": "filepicker",
                     "index": 1,
-                    "label": "Appmixer File ID",
-                    "tooltip": "Select an existing file or upload a new file from your computer to Appmixer. This file will be then uploaded to Google Drive."
+                    "label": "File ID",
+                    "tooltip": "Select an existing file or upload a new file from your computer. This file will be then uploaded to Google Drive."
                 },
                 "fileName": {
                     "type": "text",
--- a/src/appmixer/google/drive/UploadFile/component.json
+++ b/src/appmixer/google/drive/UploadFile/component.json
@@ -49,8 +49,8 @@
                 "fileId": {
                     "type": "filepicker",
                     "index": 1,
-                    "label": "File ID",
-                    "tooltip": "Select an existing file or upload a new file from your computer."
+                    "label": "Appmixer File ID",
+                    "tooltip": "Select an existing file or upload a new file from your computer to Appmixer. This file will be then uploaded to Google Drive."
                 },
                 "fileName": {
                     "type": "text",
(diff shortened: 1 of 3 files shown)


src/appmixer/google/drive/UploadFile/component.json
```diff
@@ -49,8 +49,8 @@
                 "fileId": {
                     "type": "filepicker",
                     "index": 1,
-                    "label": "File ID",
-                    "tooltip": "Select an existing file or upload a new file from your computer."
+                    "label": "Appmixer File ID",
```
vtalas: @ZbynekPelunek We should avoid mentioning "Appmixer" anywhere within a connector. When a customer uses the embedded integration, the end user isn’t familiar with Appmixer or the Appmixer File ID, so these references would be confusing.

vtalas (changes requested):

src/appmixer/google/drive/UploadFile/component.json
ZbynekPelunek: I see but it is kinda confusing now too because it is asking for a File ID, so you would think you can also map a File ID from Google Drive, which is wrong and caused this issue. That's why I wanted to distinguish it so you know what File ID is expected in this field.

vtalas: @ZbynekPelunek I understand your reasoning, and it makes sense. However, since this is a file picker, when the user clicks on it, it already indicates that it’s not a Google Drive file ID. The rule about not mentioning “Appmixer” is quite strict. Also, take a look at the other UploadFile components (SharePoint, OneDrive, Dropbox) - they also use “File ID.” Please create an issue for this so we can discuss it with the product team, as there doesn’t seem to be an easy solution from the connector side.

vtalas (approved):

PR #847, opened 2025-11-03
Figma (minor) Scopes Updated with API changes (app verification)

sayam-nasir:


vtalas (changes requested):
@sayam-nasir  this is a breaking change, as the engine won't match the access token after the update. re-authentication is needed

sayam-nasir:
@vtalas but is it a released connector? I think not then we can go ahead with the version 1.0.1 right? the only flows we have currently are the test flows?

vtalas:
@sayam-nasir  figma has been released in August. It was in the skip list - it wasn't pre-installed to the new Appmixer instances, however anyone can go the BO and install the connector and use use it with their own clientId & client secret.

src/appmixer/figma/bundle.json
```diff
@@ -1,9 +1,12 @@
 {
     "name": "appmixer.figma",
-    "version": "1.0.0",
+    "version": "1.0.1",
```
vtalas: major release - it should be 2.0.0

vtalas (changes requested):

src/appmixer/figma/bundle.json
sayam-nasir: Thanks, updated!

vtalas (approved):

PR #851, opened 2025-11-04
Brevo.SendEmail (patch) Adding template ID and template params in request

sayam-nasir:
Added template ID and template params

### DIFF
--- a/src/appmixer/brevo/bundle.json
+++ b/src/appmixer/brevo/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.brevo",
-    "version": "1.1.1",
+    "version": "1.2.0",
     "changelog": {
         "1.0.1": [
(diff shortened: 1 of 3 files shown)


src/appmixer/brevo/bundle.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.brevo",
-    "version": "1.1.0",
+    "version": "1.1.1",
```
vtalas: it's minor release -> 1.2.0

vtalas (changes requested):

src/appmixer/brevo/bundle.json
sayam-nasir: thanks, updated!

vtalas (approved):

PR #858, opened 2025-11-07
Grok AI (new)

ZbynekPelunek:


src/appmixer/ai/grok/ImageGeneration/component.json
```diff
@@ -0,0 +1,118 @@
...
+                        "label": "User ID",
+                        "tooltip": "Unique end-user identifier for monitoring and abuse detection.",
+                        "index": 3
+                    },
+                    "outputType": {
+                        "type": "select",
+                        "label": "Output Type",
+                        "tooltip": "Choose how to receive the result set: as one list, first item only, one item at a time, or stream to CSV file.",
+                        "options": [
+                            {
```
sayam-nasir, line 85: Not enough options for outputType

src/appmixer/ai/grok/core/ImageGeneration/component.json
```diff
@@ -0,0 +1,118 @@
...
+        {
+            "name": "out",
+            "source": {
+                "url": "/component/appmixer/grok/core/ImageGeneration?outPort=out",
+                "data": {
+                    "properties": {
+                        "generateOutputPortOptions": true
+                    },
+                    "messages": {
+                        "in/outputType": "inputs/in/outputType"
```
sayam-nasir: required values should go in if we want output port options displayed

src/appmixer/ai/grok/core/ChatCompletion/ChatCompletion.js
```diff
@@ -0,0 +1,110 @@
...
+
+module.exports = {
+    async receive(context) {
+
+        const {
+            model,
+            messagesRole,
+            messagesContent,
+            messagesName,
+            messagesToolCallId,
```
sayam-nasir: I think this is overcomplicating the input for chat completion (refer to other ai connectors) where will the user get this id from, if it's not required i think we can keep it as clear as possible cc: @vtalas

src/appmixer/ai/grok/core/ImageUnderstanding/ImageUnderstanding.js
```diff
@@ -0,0 +1,63 @@
...
+            method: 'POST',
+            url: 'https://api.x.ai/v1/chat/completions',
+            headers: {
+                'Authorization': `Bearer ${context.auth.apiKey}`,
+                'Content-Type': 'application/json'
+            },
+            data: requestBody
+        });
+
+        context.log({ step: 'response', data });
```
sayam-nasir: log can be removed

sayam-nasir (changes requested):

sayam-nasir (commented):
I think this should be under ai folder 

cc: @vtalas

vtalas:
> I think this should be under ai folder
> 
> cc: @vtalas

yes, definitely, it should be in the `ai` folder

src/appmixer/ai/grok/ImageGeneration/component.json
ZbynekPelunek: This is not typical "List XYZ" component but one where you set the amount of items you want it to return (max 10), in that case to me options "first item only" and "store into csv" felt unnecessary, asked @vtalas for confirmation and he agrees, so the 2 options are enough

src/appmixer/ai/grok/core/ChatCompletion/ChatCompletion.js
ZbynekPelunek: You are right, I don't know how I ended up with so many parameters, when the API docs shows only like 5. I have refactored the component so now the messages are in an array and you can enter multiple as either System, User or Assistant. No Tool option anymore

PR #862, opened 2025-11-10
refactor npm install script for connectors

jirihofman:
Fixes https://github.com/Appmixer-ai/appmixer-components/issues/2561

- update install script to fail on error, switched from bash to node
- updated langchain dependencies
- removed package.json from claude - not used at all
- added unit tests


src/appmixer/ai/claude/package.json
```diff
@@ -1,5 +0,0 @@
-{
-  "dependencies": {
-    "langchain": "0.3.24"
```
vtalas, line 3: why you've removed the package.json? langaing is used in the `TransformTextToJSON`

vtalas (approved):

PR #866, opened 2025-11-14
Copilot/add triggerdev connector

jirihofman:
Closes https://github.com/Appmixer-ai/appmixer-components/issues/2570

:copilot: https://github.com/jirihofman/appmixer-connectors/pull/58


jirihofman:
This is way off. It should focus on CRUD for Tasks. This can currently only be done via their SDK.

PR #887, opened 2025-11-28
HelpScout.FindConversations (minor) Adding status in filters

sayam-nasir:


src/appmixer/helpscout/core/FindConversations/component.json
```diff
@@ -36,10 +37,46 @@
...
+                        "options": [
+                            {
+                                "clearItem": true,
+                                "content": "-- Clear selection --"
+                            },
+                            {
+                                "label": "Active",
+                                "value": "active"
+                            },
+                            {
```
ZbynekPelunek, line 54: By default, the API returns only active conversations, so leaving this unselected doesn't return all (which I would expect when I don't use a filter)
I think it would be better if you made the select like this:
1. Remove option to clear selection
2. Add "defaultValue": "all", or "active"

```suggestion
                    "status": {
                        "type": "select",
                        "label": "Status",
                        "index": 2,
                        "tooltip": "Filter conversations by status. Defaults to active.",
                        "defaultValue": "active",
                        "options": [
                            {
                                "label": "Active",
                                "value": "active"
                            },
                            {
```

src/appmixer/helpscout/bundle.json
```diff
@@ -1,9 +1,12 @@
 {
     "name": "appmixer.helpscout",
-    "version": "1.0.1",
+    "version": "1.0.2",
     "changelog": {
         "1.0.1": [
             "Initial version"
+        ],
+        "1.0.2": [
+            "Improved query handling in Find Conversations component"
```
ZbynekPelunek: Why have a patch note for it instead of just bumping the "initial version"? Also I don't think the patch note is even true since you added a new filter option "status" and didn't touch "query" at all :D

src/appmixer/helpscout/core/FindConversations/component.json
sayam-nasir: Since the api already defaults to active we do not have to send the value "active" and the tooltip says "defaults to active" should suffice

src/appmixer/helpscout/bundle.json
sayam-nasir: Yea, I also thought about not adding a patch note but then I added because launching the connector on base version 1.0.2 sounded weird, also, these all filters are query parameters and then a parameter called "query" is separate. So in general, query handling

vtalas: @sayam-nasir it's ok to have have the initial version 1.0.2. It doesn't make sense to have release notes for each bump before the connector is released.

vtalas (approved):

PR #891, opened 2025-12-08
Cloudflare Waf 

vtalas:
A few notes for the actual solution:
- The connector should only overwrite the rules created by it, no other rules.
- Each rule in cloudflare can't exceed 4096 character capacity.
- The connector should update existing dedicated rules with new IPs until the capacity is met, and once reached, it should create a new dedicated rule.
  That way, we won't spam the customer's WAF with multiple rules without good reason and avoid future issues with their Cloudflare.


vtalas:
ticket https://appmixer.freshdesk.com/a/tickets/9440

jirihofman (approved):

vtalas:
verified on https://my.clientio.appmixer.cloud/

PR #893, opened 2025-12-09
utils.csv: improve error handling 

vtalas:
Summary of Changes to CSVProcessor.js

  addRows method fixes:

  1. Error propagation - Moved setInterval inside promise scope so errors properly reject instead of throwing into void
  2. Null reference - Added lastLine && check before calling .toString()
  3. lock.extend() error handling - Wrapped in try-catch to prevent unhandled rejections
  4. Race condition - Added destroyed flag to prevent double-cleanup
  5. Empty file handling - Added check for missing headers with clear error message
  6. IMPORTANT Improve lock extending behavior - periodically check the remaining time, extend the lock when the time to expire is less then 40 seconds

  Other method fixes:

  7. getRows - Added settled flag to prevent lock being unlocked multiple times and promise being resolved multiple times
  8. loadFile - Pipeline errors now propagate via resultStream.destroy(err) instead of being silently ignored
  9. editCSV - Removed incorrect err parameter from close event handler; refactored to use formatRow()
  10. updateRows - Renamed inner values to indexedValuesArray to avoid variable shadowing
  11. addColumn - Changed if (options.index) to if (typeof options.index === 'number') so index 0 works

  New method:

  12. escapeCell() - Added CSV escaping per RFC 4180 (quotes values containing delimiter, quotes, or newlines) to prevent CSV injection

https://appmixer.freshdesk.com/a/tickets/9441


jirihofman (approved):

src/appmixer/utils/csv/CSVProcessor.js
```diff
@@ -370,55 +378,80 @@ module.exports = class CSVProcessor {
     async addRows({ rows }) {
 
         const config = this.context.config;
+        const defaultTtl = 180000; // Default and minimum  TTL is 3 minutes
+        const extendThreshold = 40000; // Extend the lock when less than 40 seconds left
```
martin-krcmar: Should all these values be configurable via the Backoffice?

RamBachkheti:
@jirihofman Could you please let me know what I need to test using the addRows action?

vtalas:
> @jirihofman Could you please let me know what I need to test using the addRows action?

@RamBachkheti this fixed is based on customer's findings. Please do the stress tests - large files, many rows. They encountered “Cannot read properties of undefined (reading 'charCodeAt')” and Lock acquisition error with 30 retry attempts

RamBachkheti:
Verified the fix with following environment:

**Environment Info:**
Server: https://my.clientio.appmixer.cloud/
Version: 6.3.0
Connector Version: v2.0.7

Test Flow: https://my.clientio.appmixer.cloud/designer/ffd50785-9430-4db5-9259-a2dab5c8c95a/019a09c0-8aff-48fc-9daa-828c892d2f4f

PR #910, opened 2026-01-07
AWS (major): migrate aws connectors to SDK v3

jirihofman:
Implements all of https://github.com/Appmixer-ai/appmixer-components/issues/2586

- migrate from AWS SDK v2 to v3
- keep inports and outports the same as were to make it backwards compatible
- remove "service" files `aws-commons.js` and `auth.js` and also `packagejson`
    - each connector implements its own auth (same for all)
    - does not depend on a shared lib
    - see update examples below
- remove Promise dependency from S3

### Use case: Update single connector with v3 without redeploying the whole service

Updating single connector (e.g. S3 using v3) must keep working other connectors the old way (e.g. Lambda using v2). If there is a shared service-wide auth or lib in S3, updating S3 to v3 would break Lambda using v2 because `aws-commons.js` and `auth.js` would be replaced with new versions incompatible with v2.

By removing the shared service-wide files it is ensured that each connector is independent and can be updated separately without breaking others - keeping the original v2 service-wide files in the service folder.

The question is how (when) to remove the old v2 files? Should be done when all connectors are migrated to v3, but how to track that?

Instead of broken #909 with too many conflicts


src/appmixer/aws/s3/UpdatedObject/UpdatedObject.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
```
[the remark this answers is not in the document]
jirihofman: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Appmixer-ai/appmixer-connectors/pull/910#discussion_r2668846776)

vtalas (commented):
## Code Review

### Questions
1. **Auth service name change**: Changing `appmixer:aws` to `appmixer:aws:s3`, `appmixer:aws:lambda`, etc. - will existing users need to re-authenticate? This should be documented in migration notes.
2. **Body output format in GetObject**: Is the `Buffer.toJSON()` format change intentional? This returns `{ type: 'Buffer', data: [...] }` which differs from v2 output.

See inline comments for specific issues.

src/appmixer/aws/s3/NewObject/NewObject.js
```diff
@@ -33,43 +34,50 @@ module.exports = {
     async receive(context) {
 
         const { bucket } = context.properties;
-        const { s3 } = commons.init(context);
+        const { s3 } = lib.init(context);
```
vtalas, line 37: The `results` variable from `Promise.allSettled` is assigned but never used. If any record processing fails, errors will be silently swallowed.

Consider either:
1. Logging failed results: `results.filter(r => r.status === 'rejected').forEach(r => context.log('error', r.reason))`
2. Or removing the unused variable assignment if errors should be ignored

src/appmixer/aws/s3/UpdatedObject/UpdatedObject.js
```diff
@@ -33,27 +34,28 @@ module.exports = {
     async receive(context) {
 
         const { bucket } = context.properties;
-        const { s3 } = commons.init(context);
+        const { s3 } = lib.init(context);
```
vtalas, line 37: Same issue here - `results` variable is unused. Failed record processing will be silently swallowed.

src/appmixer/aws/s3/GetObject/GetObject.js
```diff
@@ -18,10 +20,30 @@ module.exports = {
...
+                const chunks = [];
+                for await (const chunk of response.Body) {
+                    chunks.push(chunk);
+                }
+                bodyBuffer = Buffer.concat(chunks);
+            }
+            const object = {
+                Bucket: bucket,
+                Key: key,
+                ...response,
```
vtalas, line 38: Using `Buffer.toJSON()` returns `{ type: 'Buffer', data: [...] }` which is different from SDK v2 output format. This could be a breaking change for existing flows.

Consider using `bodyBuffer.toString('base64')` for text-safe encoding, or document this format change.

src/appmixer/aws/lambda/Invoke/Invoke.js
```diff
@@ -22,27 +25,51 @@ module.exports = {
...
+            // In SDK v3, Payload is a Uint8Array, convert to string using Buffer for Node.js compatibility
+            const PayloadString = response.Payload
+                ? Buffer.from(response.Payload).toString('utf-8')
+                : null;
+
+            return context.sendJson({
+                ExecutedVersion: response.ExecutedVersion,
+                FunctionError: response.FunctionError,
+                LogResult: response.LogResult,
+                Payload: PayloadString,
```
vtalas, line 64: Wrapping errors loses the stack trace and error code. Consider preserving more context:
```javascript
const newError = new Error(error.message);
newError.code = error.code || error.name;
throw newError;
```

src/appmixer/aws/sns/SendSMS/SendSMS.js
```diff
@@ -11,48 +13,54 @@ module.exports = {
...
         }
 
         if (!phoneNumber) {
-            throw new context.CancelError('Phone Number. is required');
+            throw new context.CancelError('Phone Number is required');
         }
 
         if (!message) {
-            throw new context.CancelError('Message. is required');
+            throw new context.CancelError('Message is required');
```
vtalas, line 24: The message attribute names changed from `DefaultSMSType`/`SenderID` to `AWS.SNS.SMS.SMSType`/`AWS.SNS.SMS.SenderID`. This is correct for SDK v3, but verify existing SMS functionality still works as expected.

jirihofman:
@vtalas 

Ad 1) Added info that re-authentication is needed

Ad 2) This is intentional to keep the same format - even though it is a bit unfortunate. See output from current prod version

image

src/appmixer/aws/lambda/Invoke/Invoke.js
jirihofman: This is fine, we don't need stack trace nor error code. This is enough

image

src/appmixer/aws/s3/GetObject/GetObject.js
jirihofman: This is to keep the same output format, screenshot added earlier

src/appmixer/aws/s3/NewObject/NewObject.js
jirihofman: The `results` variable is used in https://github.com/Appmixer-ai/appmixer-connectors/pull/910/changes#diff-e40a1181db3d8298b9fb966fba78ee4784d6ac036f619dd881840150a4daef9dR76-R80

Exactly as you suggest in option 1

src/appmixer/aws/s3/UpdatedObject/UpdatedObject.js
jirihofman: The `results` variable is used in https://github.com/Appmixer-ai/appmixer-connectors/pull/910/changes#diff-e40a1181db3d8298b9fb966fba78ee4784d6ac036f619dd881840150a4daef9dR76-R80

Exactly as you suggest in option 1

src/appmixer/aws/sns/SendSMS/SendSMS.js
jirihofman: Verified by running the same flow with current PROD version and this version 🆗

vtalas (approved):

PR #928, opened 2026-01-16
MySQL (major): enhance security with parameterized queries, add query validation

jirihofman:
Part of https://github.com/Appmixer-ai/appmixer-components/issues/2509

Without this it runs the query. Multi queries are [blocked by default by the library](https://github.com/mysqljs/mysql?tab=readme-ov-file#multiple-statement-queries) default settings.

image

Breaking because there might be some weird stuff in the `SQL query` field.


jirihofman:
Interesting note: Don't name your branch `claude/mysql-sql-injection`, GitHub won't let you push it 🤣

vtalas (approved):

PR #957, opened 2026-02-11
Unkey (fix): add `credits` fields to `CreateKey`

jirihofman:
After clarification of the docs by Unkey in https://github.com/unkeyed/unkey/issues/4574 it is now clear how to use `credentials` field.

- Added that to `CreateKey` so that increment/decrement can be used afterwards by `UpdateKeyCredits`.
- changed `expires` to be datepicker in Inspector

Fixes https://github.com/Appmixer-ai/appmixer-components/issues/2567#issuecomment-3879084077 - Issue 1

### DIFF
--- a/src/appmixer/unkey/bundle.json
+++ b/src/appmixer/unkey/bundle.json
@@ -1,8 +1,8 @@
 {
     "name": "appmixer.unkey",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "changelog": {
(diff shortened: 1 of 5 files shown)


vtalas (changes requested):
please update the `expires` (CreateKey, UpdateKey) it should be date-time, not number

vtalas (approved):

PR #981, opened 2026-02-27
Forms: Checkbox form field does not emit value (patch)

vtalas:
https://appmixer.freshdesk.com/a/tickets/9498

### DIFF
--- a/src/appmixer/utils/forms/FormTrigger/FormTrigger.js
+++ b/src/appmixer/utils/forms/FormTrigger/FormTrigger.js
@@ -41,7 +41,15 @@ module.exports = {
                 return context.response(formPage, 200, { 'Content-Type': 'text/html' });
             } else if (context.messages.webhook.content.method === 'POST') {
                 // Form submission. Send entry to the entry output port.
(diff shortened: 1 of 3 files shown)


jirihofman (approved):
LGTM 🦀

vtalas:
@RamBachkheti  FromTrigger is ready for testing on clientIO instance

RamBachkheti:
@jirihofman : I don’t have access to Freshdesk. Could you please share more details about this issue and the fix? Thanks!

vtalas:
> @jirihofman : I don’t have access to Freshdesk. Could you please share more details about this issue and the fix? Thanks!

Small bug I just hit. In the Form trigger, I have a checkbox field type. When checked, the item does not emit any value. Should be simple to reproduce.
image

image


Result: 
image

expected is true / false

RamBachkheti:
@vtalas : Verified the fix with following environment:

**Environment Info:**
Server: Dev 
Version: v 6.3.11
Connector Version: 1.0.2
Browser: Google Chrome

Screenshot 2026-03-24 at 4 15 55 PM

PR #1007, opened 2026-03-10
postgres(minor): async query support via dblink + MongoDB job tracking

### LINKED ISSUE #977: PostgreSQL: Long-Running Query Support
# Long-Running Query Support in PostgreSQL Component

## Overview

Currently, the Appmixer PostgreSQL Query component has a ~23-minute timeout limit due to Appmixer's message timeout constraints. Queries exceeding this duration will fail.

We've evaluated solutions to enable long-running queries without blocking Appmixer's message handling. This issue summarizes the two viable approaches.

### Option 1: Using `dblink` Extension

**What it does**: `dblink` is a PostgreSQL extension that allows queries to execute asynchronously in background sessions via a remote connection mechanism. The component would:
1. Fire the query asynchronously via `dblink_send_query()`
2. Poll `dblink_get_result()` every 5 seconds to check if the query is done
3. Return results once complete

**Advantages**:
- Works on **Amazon Aurora PostgreSQL**
- Mature, widely-used extension
- Non-blocking polls don't consume Appmixer resources

**Disadvantages**:
- Requires `CREATE EXTENSION IF NOT EXISTS dblink;` on the database
- Adds ~5-10ms per poll tick (network round-trip)
- Requires additional refactoring of the connector to support persistent connections: https://github.com/Appmixer-ai/appmixer-components/issues/2518

**Installation**:
```sql
CREATE EXTENSION IF NOT EXISTS dblink;
```

---

### Option 2: Using `pg_background` Extension

**What it does**: `pg_background` launches queries in true background worker processes managed by PostgreSQL. The component would:
1. Launch the query with `pg_background_launch(sql)` which returns a PID immediately
2. Poll `pg_stat_activity` to check if the PID still exists
3. Call `pg_background_result(pid)` once the worker is done

**Advantages**:
- True PostgreSQL background workers (no separate connection needed)
- Simpler, cleaner architecture
- Minimal polling overhead
- Minimal refactoring of the current `Query` component

**Disadvantages**:
- **NOT available on Amazon Aurora** (AWS doesn't include this extension in their PostgreSQL distributions)
- Only works on self-managed PostgreSQL installations

**Installation** (if available):
```sql
CREATE EXTENSION IF NOT EXISTS pg_background;
```

---

## Recommendation

- `dblink` for wider audience
- `pg_packground` for quicker release

## Next Steps

**Questions:**
1. Are you running on Amazon Aurora, or self-managed PostgreSQL?
2. Do you have permission to install extensions on your database?
3. Would you be comfortable installing `dblink`?

Once we gather requirements, we can proceed with implementation.

---

## Technical Details (for reference)

The implementation will add a `fireAndForget` toggle to the Query component:
- When **enabled**: Query runs asynchronously, component returns immediately, results polled and returned via timeout mechanism
- When **disabled**: Standard synchronous execution (current behavior)

This keeps the component backward-compatible while adding the long-running query capability.

salatv-ai:
### Overview

Adds a non-blocking async execution path to the PostgreSQL `Query` component. Long-running queries are dispatched via `dblink_send_query` and tracked in MongoDB — allowing the Appmixer flow to return immediately and deliver results later without occupying a worker thread.

---

#### Async Execution Flow

```
asyncMode=true
  → Query.js: insertOne(pgAsyncJobs) + dblink_send_query()
  → return immediately (no blocking)

jobs.js (poll every 5s):
  → findOne({ status: "pending" })
  → dblink_is_busy() → still running? skip
  → done → dblink_get_result() → updateOne({ status: "done", result })
  → triggerComponent(flowId, componentId, _asyncJobId)

Query.js receive(_asyncJobId):
  → findOne(pgAsyncJobs, { _asyncJobId })
  → sendJson(out)
```

#### Key Components

| File | Purpose |
|---|---|
| `AsyncJobModel.js` | MongoDB model for `pgAsyncJobs` collection; TTL index for automatic cleanup |
| `connections.js` | Process-global `pg.Client` pool for persistent `dblink` sessions; survives `require()` cache clears (same pattern as Kafka connector) |
| `jobs.js` | Background poller (5s) + cleanup job (10min); distributed locks prevent duplicate processing across cluster nodes |
| `config.js` | Configurable polling intervals and TTL thresholds |
| `plugin.js` | Plugin entry point — registers background jobs on connector load |
| `Query.js` | Unified component: sync path unchanged; async path added as opt-in via `asyncMode` toggle |

#### Why `dblink`?

PostgreSQL's `dblink` extension allows one connection to fire a query and another to poll its status — decoupling execution from result delivery. This avoids long-held connections and worker starvation on slow queries.

#### Why MongoDB for job state?

Appmixer already uses MongoDB as its operational store. Storing async job state there (vs. Redis or in-memory) gives us:
- **Persistence** across restarts
- **Cluster-safe** — any node can pick up a job
- **TTL cleanup** — old jobs expire automatically

---

### Sync Path

Unchanged. `asyncMode=false` (default) goes through the existing execution path with zero overhead.


salatv-ai:
## Code Review — Postgres Async Query (PR #1007)

### ✅ What looks good
- **Architecture**: dblink-based async execution is a solid approach for long-running PG queries. Clean separation: `connections.js` (in-memory connection tracking), `AsyncJobModel.js` (Mongo persistence), `jobs.js` (polling/cleanup), `plugin.js` (init).
- **Resilience**: Good recovery pattern — if a node restarts, `jobs.js` detects orphaned jobs (no local connection) and restarts the query. Stuck job cleanup with configurable threshold is nice.
- **Distributed locking**: Both poll and cleanup jobs use locks to avoid concurrent execution across cluster nodes.
- **TTL index** on `updatedAt` for auto-cleanup of completed/errored jobs.
- **Sync mode preserved**: Existing sync behavior is untouched (just restructured for readability).
- **Bug fix**: `messages` → `message` typo in emptyResult output — good catch.

### ⚠️ Issues / Suggestions

**1. Security: `auth` credentials stored in MongoDB**
`startAsyncQuery()` persists `context.auth` (including `dbPassword`) directly into the `pgAsyncJobs` collection. This means plaintext PG passwords sit in Mongo. Consider:
- Encrypting sensitive fields before storage
- Or not persisting auth at all — instead, re-fetch it from the flow/component context during recovery

**2. `buildConnStr` is vulnerable to injection**
The `buildConnStr()` function concatenates auth values directly into a libpq connection string without escaping. If `dbHost`, `dbUser`, or `dbPassword` contain spaces or special characters, the connection string breaks or worse. Consider using `pg-connection-string` or at minimum quoting the values: `'${value}'`.

**3. Missing `query` validation in async mode**
Async mode wraps the user query in `SELECT row_to_json(t)::text AS _row FROM (${query}) t` — this is direct string interpolation. While the sync mode uses `QueryStream` (parameterized), the async path does not. This is a regression from the SQL injection fix in v2.1.0.

**4. `_asyncJobId` and `_asyncError` in component.json schema**
These fields are used internally by `triggerComponent()` but are not declared in `component.json` schema. While this works, it means:
- They could be set by a user in the UI (unintended behavior)
- Consider adding them as hidden/private fields or checking the source of the message

**5. Plugin requires `dblink` extension**
`CREATE EXTENSION IF NOT EXISTS dblink` requires superuser or `CREATE` privilege on the database. The tooltip mentions this, but the error message when it fails would be cryptic. Consider catching and providing a clearer error.

**6. `process.PG_ASYNC_CONNECTIONS` global**
Attaching to `process` object is fragile — it survives require cache clears but is unconventional. Comment mentions Kafka connector does the same, so this might be an established pattern in the codebase. Just flagging it.

**7. Test flows**
Four test flow files added — great coverage.

### 🔧 ESLint fix applied
Fixed 3 ESLint errors in `connections.js`:
- Double quotes → single quotes
- `_row` → `jsonRow` (no-underscore-dangle)
- Unnamed catch `_` → `e`

### Summary
Overall solid implementation. The main concerns are **security** (plaintext credentials in Mongo, connection string injection) and **SQL injection in async mode**. The architecture and recovery patterns are well-thought-out.

PR #1019, opened 2026-03-17
googleAds (minor): Add CSV mapping

jirihofman:
Use case: user wants to use the same CSV for `facebookbusiness` and `googleAds` connector components that upload CSV files.

### DIFF
--- a/src/appmixer/googleAds/artifacts/test/AddUsersFromCSVToUserList.test.js
+++ b/src/appmixer/googleAds/artifacts/test/AddUsersFromCSVToUserList.test.js
@@ -29,6 +29,74 @@ function emailCsv(rows) {
     return [header, ...lines].join('\n');
 }
 
+const SHARED_CSV_HEADER = 'first_name,last_name,personal_emails,personal_phone,mobile_phone,additional_personal_emails,personal_emails_validation_status,personal_emails_last_seen,linkedin_url,personal_address,personal_address_2,personal_city,personal_state,personal_zip,personal_zip4,contact_country,gender,age_range,married,children,income_range,net_worth,homeowner,job_title,job_title_normalized,job_title_last_updated,seniority_level,seniority_level_2,department,department_2,business_email,programmatic_business_emails,business_email_validation_status,business_email_last_seen,direct_number,professional_address,professional_address_2,professional_city,professional_state,professional_zip,professional_zip4,company_name,company_domain,company_phone,company_sic,company_naics,company_address,company_city,company_state,company_zip,company_country,company_revenue,company_employee_count,primary_industry,company_last_updated,last_updated';
+
+function sharedAudienceCsv(overrides = {}) {
+    const columns = SHARED_CSV_HEADER.split(',');
+    const defaults = {
+        first_name: 'Avery',
+        last_name: 'Stone',
+        personal_emails: 'avery.stone@example.test',
+        personal_phone: '+12025550101',
+        mobile_phone: '+12025550102',
+        additional_personal_emails: 'avery.alt@example.test',
+        personal_emails_validation_status: 'Valid',
+        personal_emails_last_seen: '2026-01-15 19:00:00',
+        linkedin_url: 'https://linkedin.example.test/in/avery-stone',
+        personal_address: '101 Cedar Ave',
+        personal_city: 'Springfield',
+        personal_state: 'IL',
+        personal_zip: '62704',
+        contact_country: 'US',
+        gender: 'X',
+        age_range: '35-44',
+        married: 'N',
+        homeowner: 'Y',
+        job_title: 'Founder',
+        job_title_normalized: 'founder',
+        job_title_last_updated: '2026-01-10 08:00:00',
+        seniority_level: 'Cxo',
+        seniority_level_2: 'owner',
+        department: 'Operations',
+        department_2: 'Leadership',
+        business_email: 'avery@work.example.test',
+        programmatic_business_emails: '\\N',
+        business_email_validation_status: 'Valid',
+        business_email_last_seen: '2026-01-15 19:00:00',
+        direct_number: '+12025550103',
+        professional_address: '200 Market St',
+        professional_city: 'Chicago',
+        professional_state: 'IL',
+        professional_zip: '60601',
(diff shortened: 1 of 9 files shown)


salatv-ai (commented):
## Review

**+750 / -233 | 8 files | CI ✅ pass**

### 👍 What's good

- **Clean refactor** — duplicated `buildUserIdentifiers` logic from both components (`Add` and `Remove`) moved into shared `lib.js` as `buildUserIdentifiersFromCsvRow`. This is the right call.
- **Explicit schema mapping** instead of magic header-name guessing. Much more robust for the use case where the same CSV is shared between Google Ads and Facebook.
- **Tests are thorough** — schema mapping coverage, case-insensitive matching, shared Facebook-style CSV, arbitrary headers, validation that schema is required. Well done.
- **`CSV_NULL_TOKENS`** — clean handling of `\\N`, `NULL`, empty values.
- **`splitMultiValue`** — nice touch, allows multiple emails/phones in a single column separated by comma/semicolon.
- **Version bump** 1.1.0 → 1.2.0 + changelog — good.

### ⚠️ Minor issues

1. **Schema validation in `buildCsvSchemaConfig`** — if an invalid `googleAdsType` comes in (e.g. typo `"emal"` instead of `"email"`), it silently gets ignored and the row is skipped with no warning. Consider validating against `CSV_MAPPING_TYPES` or logging a warning.

2. **Duplicated test helpers** — `sharedAudienceCsv()`, `googleAdsSchema()`, `SHARED_CSV_HEADER`, `BASIC_EMAIL_SCHEMA` are copy-paste identical in both test files. Would be cleaner to extract into `test/helpers.js`. Not a blocker, but tech debt.

3. **Schema inspector `levels: ["ADD"]`** — in both `component.json` files the schema uses `"levels": ["ADD"]`. For `RemoveUsersInCSVFromUserList` it's a bit confusing that the key is called `ADD` even for a remove operation. Consider renaming to something neutral or at least adding a note in the tooltip.

### 🔴 One real issue — see inline comment on bundle.json

src/appmixer/googleAds/bundle.json
```diff
@@ -1,12 +1,12 @@
...
-    "version": "1.1.0",
+    "version": "1.2.0",
     "changelog": {
         "1.0.0": [
             "Initial Google Ads release with user list and Customer Match workflow components."
         ],
-        "1.1.0": [
-            "Added support for uploading and removing users to Google Ads user lists from CSV files via two new components."
+        "1.2.0": [
+            "Added support for uploading and removing users to Google Ads user lists from CSV files via two new components using explicit schema mapping."
```
salatv-ai: 🔴 **Changelog entry for 1.1.0 was deleted.** The old record should be preserved:
```json
"changelog": {
    "1.0.0": ["Initial Google Ads release with user list and Customer Match workflow components."],
    "1.1.0": ["Added support for uploading and removing users to Google Ads user lists from CSV files via two new components."],
    "1.2.0": ["Added explicit schema mapping for CSV uploads."]
}
```

src/appmixer/googleAds/lib.js
```diff
@@ -61,6 +73,138 @@ function hashSha256(value) {
...
+function buildCsvSchemaConfig(schema) {
+    const config = {};
+    const mappings = Array.isArray(schema?.ADD) ? schema.ADD : [];
+
+    for (const mapping of mappings) {
+        const header = normalizeRowKey(mapping?.csvHeader);
+        const googleAdsType = String(mapping?.googleAdsType || '').trim();
+
+        if (!header || !googleAdsType) {
+            continue;
```
salatv-ai, line 96: ⚠️ If an invalid `googleAdsType` comes in (e.g. typo `"emal"` instead of `"email"`), it silently gets accepted into the config but will never match anything in `buildUserIdentifiersFromCsvRow`, causing rows to be skipped without warning. Consider validating against `CSV_MAPPING_TYPES`:
```js
if (!CSV_MAPPING_TYPES[googleAdsType]) {
    continue; // or throw
}
```

src/appmixer/googleAds/artifacts/test/AddUsersFromCSVToUserList.test.js
```diff
@@ -29,6 +29,74 @@ function emailCsv(rows) {
...
+        company_country: 'US',
+        company_revenue: '5000000',
+        company_employee_count: '25',
+        primary_industry: 'Software',
+        company_last_updated: '2026-01-12 10:00:00',
+        last_updated: '2026-01-15 19:00:00'
+    };
+
+    const row = columns.map(column => overrides[column] ?? defaults[column] ?? '');
+    return [SHARED_CSV_HEADER, row.join(',')].join('\n');
```
salatv-ai: 💡 `sharedAudienceCsv()`, `googleAdsSchema()`, `SHARED_CSV_HEADER`, `BASIC_EMAIL_SCHEMA` are copy-paste identical in both test files. Consider extracting to a shared `test/helpers.js`.

src/appmixer/googleAds/core/RemoveUsersInCSVFromUserList/component.json
```diff
@@ -83,18 +87,47 @@
                         "tooltip": "Numeric ID of the Customer Match user list to remove users from.",
                         "index": 5
                     },
+                    "schema": {
+                        "type": "expression",
+                        "levels": ["ADD"],
```
salatv-ai, line 92: ⚠️ `"levels": ["ADD"]` — the schema key is called `ADD` even though this is the **Remove** component. This is confusing for users configuring the component. Consider using a neutral name like `"MAPPINGS"` or at least adding a note in the tooltip explaining why it says ADD.

salatv-ai (commented):
Inline comments below 👇

salatv-ai (approved):
All comments addressed ✅

- Changelog preserved (1.1.0 + 1.2.0)
- `googleAdsType` validation with `CancelError` + test coverage
- Shared test helpers extracted to `test/helpers.js`
- `levels: ["ADD"]` convention acknowledged — consistent with the rest of the codebase

LGTM 👍

vtalas (approved):

PR #1020, opened 2026-03-18
ai.openai: Fix ioredis sentinel connection configuration

martin-krcmar:
## Summary
- Parse sentinel hosts from `"host:port"` string format to object array `[{host, port}]`
- Add support for `REDIS_PASSWORD` environment variable
- Properly separate `password` (Redis auth) and `sentinelPassword` (Sentinel auth)
- Include TLS options in sentinel mode configuration

## Context
This fix aligns with the changes made in appmixer-core PR #3563 to ensure proper connection to Redis in sentinel mode.

## Changes
The `connectRedis` function in `src/appmixer/ai/openai/lib.js` now:
1. Properly parses the comma-separated sentinel string into an array of `{host, port}` objects
2. Uses `REDIS_PASSWORD` for Redis authentication
3. Uses `REDIS_SENTINEL_PASSWORD` for Sentinel authentication  
4. Applies TLS configuration to sentinel connections

## Test plan
- [ ] Test Redis connection in sentinel mode with password authentication
- [ ] Test Redis connection in sentinel mode with TLS enabled
- [ ] Verify backward compatibility with standalone Redis mode

Fixes: https://github.com/Appmixer-ai/appmixer-components/issues/2656

### DIFF
--- a/src/appmixer/ai/openai/lib.js
+++ b/src/appmixer/ai/openai/lib.js
@@ -84,7 +84,8 @@ module.exports = {
             mode: process.env.REDIS_MODE || 'standalone',
             sentinels: process.env.REDIS_SENTINELS,
             sentinelMasterName: process.env.REDIS_SENTINEL_MASTER_NAME,
-            sentinelRedisPassword: process.env.REDIS_SENTINEL_PASSWORD,
+            password: process.env.REDIS_PASSWORD,
+            sentinelPassword: process.env.REDIS_SENTINEL_PASSWORD,
             enableTLSForSentinelMode: process.env.REDIS_SENTINEL_ENABLE_TLS,
             caPath: process.env.REDIS_CA_PATH,
             useSSL: process.env.REDIS_USE_SSL === 'true' || parseInt(process.env.REDIS_USE_SSL) > 0
@@ -101,14 +102,19 @@ module.exports = {
 
         if (connection.mode === 'replica' && connection.sentinels) {
 
-            const sentinelsArray = connection.sentinels.split(',');
+            const sentinelsArray = connection.sentinels.split(',').map(sentinel => {
+                const [host, port] = sentinel.split(':');
+                return { host, port: parseInt(port) };
+            });
 
             client = new Redis({
                 sentinels: sentinelsArray,
                 name: connection.sentinelMasterName,
-                ...(connection.sentinelRedisPassword ? { password: connection.sentinelRedisPassword } : {}),
+                ...(connection.password ? { password: connection.password } : {}),
+                ...(connection.sentinelPassword ? { sentinelPassword: connection.sentinelPassword } : {}),
                 ...(connection.enableTLSForSentinelMode ?
-                    { enableTLSForSentinelMode: connection.enableTLSForSentinelMode } : {})
+                    { enableTLSForSentinelMode: connection.enableTLSForSentinelMode } : {}),
+                ...options
             });
         } else {
             client = connection.uri ? new Redis(connection.uri, options) : new Redis();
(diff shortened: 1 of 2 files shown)


salatv-ai (commented):
Overall looks good — the separation of `password` vs `sentinelPassword` and the sentinel host parsing fix make sense. A couple of things to address below.

src/appmixer/ai/openai/lib.js
```diff
@@ -84,7 +84,8 @@ module.exports = {
             mode: process.env.REDIS_MODE || 'standalone',
             sentinels: process.env.REDIS_SENTINELS,
             sentinelMasterName: process.env.REDIS_SENTINEL_MASTER_NAME,
-            sentinelRedisPassword: process.env.REDIS_SENTINEL_PASSWORD,
+            password: process.env.REDIS_PASSWORD,
+            sentinelPassword: process.env.REDIS_SENTINEL_PASSWORD,
             enableTLSForSentinelMode: process.env.REDIS_SENTINEL_ENABLE_TLS,
             caPath: process.env.REDIS_CA_PATH,
             useSSL: process.env.REDIS_USE_SSL === 'true' || parseInt(process.env.REDIS_USE_SSL) > 0
```
salatv-ai, line 91: The sentinel host parsing doesn't handle edge cases — if a sentinel string doesn't contain `:` (e.g. just `"redis-sentinel"`), `port` will be `undefined` and `parseInt(undefined)` returns `NaN`. Consider defaulting to the standard Sentinel port 26379:

```js
const sentinelsArray = connection.sentinels.split(",").map(sentinel => {
    const [host, port] = sentinel.trim().split(":");
    return { host, port: port ? parseInt(port) : 26379 };
});
```

Also worth adding `.trim()` in case there are spaces after commas in the env var.

src/appmixer/ai/openai/lib.js
```diff
@@ -101,14 +102,19 @@ module.exports = {
 
         if (connection.mode === 'replica' && connection.sentinels) {
 
-            const sentinelsArray = connection.sentinels.split(',');
+            const sentinelsArray = connection.sentinels.split(',').map(sentinel => {
+                const [host, port] = sentinel.split(':');
```
salatv-ai: Nice catch adding `...options` here — sentinel mode was previously missing TLS options that standalone mode got. Just double-check that properties explicitly set above (like `password`, `sentinels`, `name`) won't get overwritten by anything in `options`. Since the spread is last, it would win in case of conflict. Might be safer to put `...options` first:

```js
client = new Redis({
    ...options,
    sentinels: sentinelsArray,
    name: connection.sentinelMasterName,
    ...(connection.password ? { password: connection.password } : {}),
    ...(connection.sentinelPassword ? { sentinelPassword: connection.sentinelPassword } : {}),
    ...(connection.enableTLSForSentinelMode
        ? { enableTLSForSentinelMode: connection.enableTLSForSentinelMode } : {}),
});
```

salatv-ai:
⚠️ **Missing backward compatibility for `REDIS_SENTINEL_PASSWORD`**

When only `REDIS_SENTINEL_PASSWORD` is defined (without `REDIS_PASSWORD`), the current change breaks existing setups — `password` will be `undefined` and Redis master auth fails.

Please add a fallback chain before the `new Redis()` call:

```js
// Determine passwords for Redis master and Sentinel nodes
// Priority: use specific password if available, otherwise use the general password,
// then fall back to sentinelRedisPassword
const redisPassword = connection.password || connection.sentinelRedisPassword;
const sentinelPassword = connection.sentinelRedisPassword || connection.password;
```

And keep reading `sentinelRedisPassword` from `REDIS_SENTINEL_PASSWORD` in the connection object, then use:
```js
...(redisPassword ? { password: redisPassword } : {}),
...(sentinelPassword ? { sentinelPassword } : {}),
```

See PR #1021 for the full implementation.

vtalas (approved):

vtalas (approved):

PR #1033, opened 2026-03-30
utils.converters (patch) - Fix crashing XLSX2CSV with invalid file

jirihofman:
https://appmixer.freshdesk.com/a/tickets/9566
Closes https://github.com/Appmixer-ai/appmixer-components/issues/2663

> The crash was caused by feeding a malformed XLSX container into SheetJS through [xlsx-helpers.js](vscode-file://vscode-app/Users/jirihofman/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). An [.xlsx](vscode-file://vscode-app/Users/jirihofman/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) file is just a ZIP archive with Office XML parts inside; the Boots file was not structurally valid as a ZIP workbook.

**What was wrong with the file:**
> unzip reported that the archive was missing 741111411 bytes and could not seek to the beginning of the ZIP footer. That means the file’s ZIP central directory was broken or truncated. In practice, the workbook data was corrupted, not just “an unusual XLSX”.

**Why that crashed instead of throwing a normal error:**
> With this corrupted archive, the parser did not fail cleanly in this environment and the Node process was killed with exit code 137, which is consistent with the OS terminating it during pathological parsing, typically due to memory pressure or runaway work. The root cause was the invalid file, not the converter logic for valid spreadsheets.

The fix adds a cheap ZIP sanity check before parsing, so corrupted XLSX-like files now fail fast with a clear error instead of reaching SheetJS.

### DIFF
--- a/src/appmixer/utils/converters/bundle.json
+++ b/src/appmixer/utils/converters/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.utils.converters",
-    "version": "1.4.0",
+    "version": "1.4.1",
     "changelog": {
         "1.1.0": [
             "FileId inputs changed from 'text' to 'filepicker'. Requires Appmixer 4.4.4 or higher."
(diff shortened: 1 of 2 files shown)


vtalas (approved):

vtalas:
verified on QA with testflow https://github.com/Appmixer-ai/appmixer-connectors/pull/1037

PR #1044, opened 2026-04-02
Freshdesk: Add CRUD components for Contacts

apx-vero:
Adds new contact management components for the Freshdesk connector:

## New Components
- **CreateContact** — create a new contact
- **GetContact** — get a contact by ID (full record)
- **UpdateContact** — update an existing contact
- **DeleteContact** — delete a contact
- **FindContacts** — search/filter contacts with:
  - Search by Name (autocomplete endpoint, prefix match)
  - Filter by Fields (search API with keyword syntax)
  - Email, phone, mobile, company, tag, updated-since filters
  - `outputType` select (first / all at once / one at a time / file)

## E2E Test Flow
- `test-flow-contact-lifecycle.json` — covers Create → Find → Update → Delete


DavidDurman (approved):

apx-vero:
All Copilot review comments have been addressed in the latest commit ([6366ae67](https://github.com/apx-vero/apmmixer-connectors/commit/6366ae67)):

- **`trimUndefined()`** — fixed to preserve `false`/`0` values (was using `if (body[key])` which dropped all falsy, including `view_all_tickets: false` and `private: false`)
- **GetContact + GetCompany** — axios 404 is now caught explicitly and routed to the `notFound` port (previously unreachable since axios throws on 4xx)
- **NewTicket / UpdatedTicket / DeletedTicket** — first-run baseline behavior: on the initial tick, cursor is saved but nothing is emitted (prevents flooding pre-existing records)
- **UpdatedTicket / DeletedTicket** — cursor is now always persisted even when no results are returned, preventing gaps when polling interval exceeds the lookback window
- **FindContacts + FindCompanies** — `sendArrayOutput` and `toCsv` helpers extracted to shared `lib.js`, eliminating duplication
- **FindContacts/component.json** — `updatedSince` now uses `"format": "date-time"` in schema and `"type": "date-time"` in inspector
- **bundle.json** — changelog corrected: `sourceInfo` removed from the ListTickets entry (it's not emitted there; only in NewTicket/UpdatedTicket/GetTicket)

vtalas (approved):

PR #1061, opened 2026-04-14
facebookbusiness: cache dynamic input calls, suppress errors for isSource

vtalas:
## Summary

- Adds `callEndpointCached()` to `lib.js` — SHA-256 cache key (URL + access token), `context.lock` + `context.staticCache`, configurable TTL via `config.listCacheTTL` (default 120s)
- `GetAdAccounts` and `GetCustomAudiences` use the cache only when called as a dynamic source (`isSource: true`) — normal flow calls bypass cache
- Errors during `isSource` calls return an empty array instead of propagating to the UI (no irrelevant error popups)
- All `source` definitions referencing `GetAdAccounts#toSelectArray` and `GetCustomAudiences#toSelectArray` updated with `"properties": { "isSource": true }`

## Note

This uses `isSource` (same as monday connector). The `copilot-instructions.md` documents `variableFetch` as the standard name for the same concept — these two should be aligned in a follow-up.

## Test plan
- [ ] Open inspector for CreateAudience / AddMemberToAudience — dropdown populates correctly
- [ ] Verify repeated dropdown opens do not trigger repeated API calls (cache hit)
- [ ] Revoke token / invalid auth — dropdown returns empty silently, no UI error
- [ ] Normal flow execution (no isSource) — errors propagate normally

### DIFF
--- a/src/appmixer/facebookbusiness/lib.js
+++ b/src/appmixer/facebookbusiness/lib.js
@@ -4,7 +4,39 @@ const workerpool = require('workerpool');
 
 const pool = workerpool.pool(__dirname + '/prepareMembers.js');
 
+function getCacheKey(obj) {
+    return crypto.createHash('sha256').update(JSON.stringify(obj)).digest('hex');
+}
+
+async function callEndpointCached(context, url) {
+
+    let lock;
+    try {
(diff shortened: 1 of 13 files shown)


apx-vero:
Closing — recreating under correct account (apx-vero).

PR #1073, opened 2026-04-23
feat: add MakeApiCall component to all API key connectors

apx-vero:
Implements MakeApiCall component for all connectors using apiKey authentication. Follows the standard spec from #1459 with method, url, headers, parameters, and body inputs.


apx-vero:
Closing as duplicate of #1074 — same work, but #1074 is properly rebased on `upstream/dev`.

PR #1075, opened 2026-04-23
fix: update existing MakeApiCall components to match standard spec

apx-vero:
## Summary

Updates 7 existing MakeApiCall components on API key auth connectors to comply with the standard defined in Appmixer-ai/appmixer-components#1459.

### axiom, clerk, convex, supabaseManagement, vercel
- Added missing `headers` input (textarea, JSON object)
- Added missing `parameters` input (textarea, JSON object for query params)
- Fixed `method` field to use `select` type with GET/POST/PUT/PATCH/DELETE options

### mollie
- Renamed `path` → `url` to match standard naming
- Added missing `parameters` input
- Fixed output format from `{response}` to `{status, headers, body}`
- Fixed base URL handling for relative paths

### stripe
- Renamed `customEndpoint` → `url` to match standard naming
- Added `headers` and `parameters` inputs
- Fixed output format from `{response}` to `{status, headers, body}`
- Rewrote request logic to use proper JSON body handling

## Version bumps
All 7 connectors: **patch** bump in `bundle.json`

### DIFF
--- a/src/appmixer/stripe/core/MakeAPICall/component.json
+++ b/src/appmixer/stripe/core/MakeAPICall/component.json
@@ -9,7 +9,10 @@
     },
     "quota": {
         "manager": "appmixer:stripe",
-        "resources": "requests"
+        "resources": "requests",
+        "scope": {
+            "userId": "{{userId}}"
+        }
     },
     "inPorts": [
         {
@@ -17,10 +20,13 @@
             "schema": {
                 "type": "object",
                 "properties": {
+                    "method": {
+                        "type": "string"
+                    },
                     "customEndpoint": {
                         "type": "string"
                     },
-                    "method": {
+                    "headers": {
                         "type": "string"
                     },
                     "parameters": {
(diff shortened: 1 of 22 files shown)


src/appmixer/stripe/core/MakeAPICall/component.json
```diff
@@ -9,48 +9,83 @@
...
+        }
     },
     "inPorts": [
         {
             "name": "in",
             "schema": {
                 "type": "object",
                 "properties": {
-                    "customEndpoint": {
+                    "method": {
```
vtalas, line 23: this would cause a breaking change - just rename the labels/toolptips.

vtalas (changes requested):
changes are minor, x.1.x -> x.2.0

src/appmixer/stripe/core/MakeAPICall/component.json
apx-vero: Good catch — reverted. Kept the original field names (`customEndpoint` for stripe, `path` for mollie) and only updated labels/tooltips. Also fixed version bumps to minor (→ x.2.0 / x.1.0) across all 7 connectors.

vtalas (approved):

PR #1081, opened 2026-04-28
fix(makeapicall): standardize all MakeApiCall components - key-value inputs, URL slash guard, missing fields

apx-vero:
## Summary

Standardizes all 12 MakeApiCall components in the `dev` branch.

### 1. key-value inputs (headers, parameters, body)
All three fields converted from `textarea` → `key-value` type with Key/Value/Add item/Remove item labels.
Previously missing fields (`headers`, `parameters`) added where absent.
inPorts schema updated: `string` → `array` for these fields.

### 2. JS updated to handle key-value format
Added `kvToObj(arr)` helper to convert `[{key, value}]` arrays to plain objects.
Headers spread into request headers, params set as `requestOptions.params`, body set as `requestOptions.data`.

### 3. URL slash guard
For connectors that construct `baseUrl + path` (vercel, convex, axiom):
```js
const path = url.startsWith('/') ? url : `/${url}`;
const targetUrl = /^https?:\/\//i.test(url) ? url : `https://api.example.com${path}`;
```

### 4. microsoft/sharepoint/common.js updated
Accepts `extraHeaders`, `queryParams`, `bodyData` objects instead of raw body string.

### 5. Version bumps
Minor version bump for all 12 affected connectors.

## Affected connectors
axiom, canva, clerk, convex, github, google/gmail, microsoft/dynamics, microsoft/sharepoint, notion, pinterest, supabaseManagement, vercel

### DIFF
--- a/src/appmixer/activecampaign/bundle.json
+++ b/src/appmixer/activecampaign/bundle.json
@@ -27,7 +27,7 @@
             "add MakeApiCall component"
         ],
         "1.2.0": [
-            "MakeApiCall: standardize key-value pair inputs, add URL trailing slash guard, improved error handling"
+            "MakeApiCall: standardize key-value pair inputs for headers and parameters (body remains JSON textarea), add URL trailing slash guard, improved error handling"
         ]
     }
 }
--- a/src/appmixer/activecampaign/bundle.json
+++ b/src/appmixer/activecampaign/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.activecampaign",
-    "version": "1.2.0",
+    "version": "2.0.0",
     "changelog": {
         "1.0.0": [
             "Initial version"
@@ -26,8 +26,8 @@
         "1.1.0": [
             "add MakeApiCall component"
         ],
-        "1.2.0": [
-            "MakeApiCall: standardize key-value pair inputs for headers and parameters (body remains JSON textarea), add URL trailing slash guard, improved error handling"
+        "2.0.0": [
+            "(breaking change) MakeApiCall: headers and parameters changed to key-value array inputs per standard #1459 — existing flows that previously set these as JSON strings must migrate to the key-value editor. Body remains textarea/JSON."
         ]
     }
 }
--- a/src/appmixer/activecampaign/bundle.json
+++ b/src/appmixer/activecampaign/bundle.json
@@ -1,6 +1,6 @@
 {
     "name": "appmixer.activecampaign",
-    "version": "1.1.0",
+    "version": "1.2.0",
     "changelog": {
         "1.0.0": [
             "Initial version"
--- a/src/appmixer/activecampaign/bundle.json
+++ b/src/appmixer/activecampaign/bundle.json
@@ -25,6 +25,9 @@
         ],
         "1.1.0": [
             "add MakeApiCall component"
+        ],
+        "1.2.0": [
+            "MakeApiCall: standardize key-value pair inputs, add URL trailing slash guard, improved error handling"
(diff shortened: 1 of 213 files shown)


src/appmixer/azuredocumentintelligence/core/MakeApiCall/component.json
```diff
@@ -86,22 +86,34 @@
                         ]
                     },
                     "headers": {
-                        "type": "textarea",
-                        "index": 3,
+                        "type": "key-value",
```
vtalas, line 89: it's a breaking change -it's different type of  of the input. update the bundle accordingly

vtalas: identify breaking changes in all connectors and update the bundle.json

apx-vero: Addressed in commit 7b31caac — identified all connectors with pre-existing MakeApiCall components and bumped each to their next major version:

- axiom: 1.3.0 → 2.0.0
- canva: 1.1.0 → 2.0.0
- clerk: 1.3.0 → 2.0.0
- convex: 1.3.0 → 2.0.0
- mollie: 1.1.0 → 2.0.0
- notion: 1.1.0 → 2.0.0
- pinterest: 1.1.0 → 2.0.0
- stripe: 1.2.0 → 2.0.0
- supabaseManagement: 1.2.0 → 2.0.0
- vercel: 1.3.0 → 2.0.0
- github: 2.2.0 → 3.0.0
- google/gmail: 2.1.0 → 3.0.0
- microsoft/dynamics: 1.1.0 → 2.0.0
- microsoft/sharepoint: 1.4.0 → 2.0.0
- azuredocumentintelligence: 2.2.0 → 3.0.0

All changelog entries updated with `(breaking change)` annotation.

apx-vero (commented):
Addressed Copilot feedback in commit efced71b:

1. **Breaking change version bumps** — vapi, virustotal, xtremepush, webflow, voys, verifyemail, and userengage all bumped to major versions (2.0.0 / 3.0.0 as appropriate), matching the convention established by vercel/axiom/etc.

2. **kvToObj empty-key filtering** — updated in all 67 MakeApiCall.js files to filter out rows where `key` is falsy/empty before building the object: `arr.filter(({ key }) => key && key.trim() !== '')`

3. **Changelog wording** — replaced 'improved error handling' with accurate description across all affected connectors: '(breaking change) MakeApiCall: standardize key-value pair inputs for body, headers, and parameters — existing flows using string/textarea input must be updated'

4. **Merge conflicts** — resolved all 44 conflicting bundle.json files (upstream dev added output-examples version bumps; both sets of changelog entries are preserved, version set to the higher of the two).

Re: the body/key-value concern — keeping body as key-value intentionally for now (consistent with headers/params), noting this limits nested JSON payloads. Can revisit if needed (e.g. body stays as textarea while headers/params use key-value).

src/appmixer/activecampaign/bundle.json
```diff
@@ -27,7 +27,7 @@
             "add MakeApiCall component"
         ],
         "1.2.0": [
-            "MakeApiCall: standardize key-value pair inputs, add URL trailing slash guard, improved error handling"
+            "(breaking change) MakeApiCall: standardize key-value pair inputs for body, headers, and parameters — existing flows using string/textarea input must be updated"
         ]
```
vtalas, line 31: if this is a breaking change, the new version should be 2.0.0 then

vtalas (approved):

PR #1123, opened 2026-06-01
hubspot(minor): add ContactPropertyChanged, NewContactInList, FormSubmission triggers

apx-vero:
## Summary

Implements [Appmixer-ai/appmixer-components#2722](https://github.com/Appmixer-ai/appmixer-components/issues/2722)

### New components

**`ContactPropertyChanged`** (webhook trigger)
- Fires when a specific contact property changes value
- Uses the existing `contact.propertyChange` HubSpot webhook subscription
- Filters in `receive()` to only emit for the user-configured `propertyName`
- `routes.js` updated to dynamically include a HubSpot subscription for custom properties via listener params

**`NewContactInList`** (polling trigger, 5 min interval)
- Fires when a new contact is added to a specific HubSpot list
- Polls `/crm/v3/lists/{listId}/memberships/join-order`
- First poll seeds the seen-set without emitting; subsequent polls emit only new members

**`FormSubmission`** (polling trigger, 5 min interval)
- Fires when a new form is submitted in HubSpot
- Polls `/form-integrations/v1/submissions/by-form/{formGuid}`
- Tracks last-seen submission timestamp in state; emits in chronological order

### Other changes
- `routes.js`: `getSubscriptionsByType` now accepts `params` and includes `params.propertyName` in contact propertyChange subscriptions to support custom properties
- `bundle.json`: version bumped `4.5.0 → 4.6.0`

## Checklist
- [x] Validation passes (`node scripts/validate.js`)
- [x] ESLint clean
- [x] Bundle version bumped (minor)


src/appmixer/hubspot/routes.js
```diff
@@ -23,7 +23,7 @@ module.exports = async (context) => {
 
             try {
                 const subscriptionType = eventName.split(':')[0];
-                const subscriptions = getSubscriptionsByType(subscriptionType, context);
+                const subscriptions = getSubscriptionsByType(subscriptionType, context, params);
                 const results = await getHubSpotSubscriptions(context, params);
                 const currentActiveSubs = results.filter(s => s.enabled).map(s => s.eventType);
```
apx-vero: **🔴 Custom-property subscription is never created on any non-virgin portal — the headline feature silently does nothing.**

Reconciliation compares only `eventType`: `currentActiveSubs` is `results.filter(s => s.enabled).map(s => s.eventType)`, and `getSubscriptionsByType` now emits one entry **per propertyName**, all sharing eventType `contact.propertyChange`. Once *any* enabled propertyChange subscription exists (guaranteed after `UpdatedContact` has ever started — subscriptions are app-wide), every candidate passes the `includes` check, `subscriptionsToCreate` stays empty, and the handler returns. HubSpot never subscribes to the custom property; the flow starts cleanly and the trigger never fires.

The disabled-sub activation path below is broken the same way: `results.find(r => r.eventType === sub.subscriptionDetails.subscriptionType)` matches by eventType only, PATCHes the **first** hit and then `return`s — `createHubSpotSubscriptions` is never reached for the remaining entries, including the one property the listener actually needs.

Fix: diff desired vs. actual on **(eventType, propertyName)** pairs, and don't `return` after activating one subscription.

src/appmixer/hubspot/routes.js
```diff
@@ -23,7 +23,7 @@ module.exports = async (context) => {
 
             try {
                 const subscriptionType = eventName.split(':')[0];
-                const subscriptions = getSubscriptionsByType(subscriptionType, context);
+                const subscriptions = getSubscriptionsByType(subscriptionType, context, params);
```
apx-vero: **🔴 Concurrent listener registrations silently drop a property's subscription — permanently.**

The outer `catch (err) { if (err.message !== 'locked') throw err; }` swallows lock contention without retry. Pre-PR that was safe: every listener of a subscriptionType requested the same constant set, so whichever registration won produced the right outcome. Now each `ContactPropertyChanged` listener contributes a **distinct** `params.propertyName`. Two flows watching properties X and Y starting at the same time (e.g. tenant restart re-adds listeners) race on `contact.propertyChange:{portalId}`; the loser's `propertyName` is never ensured and there is no reconciliation path that ever retries it. That trigger is permanently dead with zero error surfaced.

Also an altitude note: this ensures subscriptions from **one** listener's params at registration time. Since subscriptions are app-wide shared state, the durable design is to reconcile from *all* registered listeners' params (and retry on lock contention).

src/appmixer/hubspot/crm/NewContactInList/NewContactInList.js
```diff
@@ -0,0 +1,96 @@
...
+
+        if (!context.messages.timeout) {
+            return;
+        }
+
+        const hubspot = new Hubspot(context.auth.accessToken);
+        const { listId, properties } = context.properties;
+        const state = context.state || {};
+        const seenRecordIds = new Set(state.seenRecordIds || []);
+        const isFirstRun = !state.seenRecordIds;
```
apx-vero: **🔴 `isFirstRun` is always `false` — the first poll emits the entire list.**

`start()` saves `seenRecordIds: []`, and `![]` is `false`, so this check never fires and the seed-without-emitting branch below is dead code. On the first poll the seen-set is empty and **every existing list member** is batch-read and emitted as "new" — on a 5k-contact list that's 5k spurious trigger fires flooding downstream actions, directly contradicting the comment in `start()`.

`FormSubmission` in this same PR stores an explicit `isFirstRun: true` flag — do the same here (or don't save `seenRecordIds` in `start()` so this check is genuinely true on first run).

src/appmixer/hubspot/crm/FormSubmission/FormSubmission.js
```diff
@@ -0,0 +1,95 @@
...
+                    params
+                );
+
+                const results = data.results || [];
+                after = data.paging?.next?.after;
+
+                for (const submission of results) {
+                    const submittedAt = submission.submittedAt || 0;
+                    if (submittedAt > lastSeenAt) {
+                        if (!isFirstRun) {
```
apx-vero: **🔴 Submissions arriving between flow start and the first poll are permanently lost.**

Flow starts at T0 (`lastSeenAt=T0`, `isFirstRun=true`). A form is submitted at T0+2min. First poll at T0+5min: `submittedAt > lastSeenAt` is true, but this `!isFirstRun` guard skips the push — while `newLastSeenAt` still advances to T0+2min and is saved. Every later poll excludes the submission (`submittedAt <= lastSeenAt`). Silently swallowed, never emitted.

The flag is redundant: `lastSeenAt = Date.now()` in `start()` already excludes historical submissions. Its only observable effect is dropping up to one full poll interval of real submissions. Delete `isFirstRun` entirely.

src/appmixer/hubspot/crm/NewContactInList/NewContactInList.js
```diff
@@ -0,0 +1,96 @@
...
+            } while (after);
+        } catch (err) {
+            await context.log({ step: 'hubspot-new-contact-in-list-poll-error', listId, error: err.message });
+            // Save state and reschedule even on error.
+            await context.saveState({ seenRecordIds: Array.from(seenRecordIds) });
+            return context.setTimeout({}, context.config?.pollIntervalMs || DEFAULT_POLL_INTERVAL_MS);
+        }
+
+        // Save updated seen set.
+        await context.saveState({ seenRecordIds: Array.from(seenRecordIds) });
```
apx-vero: **🔴 State is saved (IDs marked seen) *before* the un-chunked batch/read — a failure here permanently swallows contacts AND kills the trigger.**

Sequence: `saveState` marks all `newRecordIds` as seen → `batch/read` below runs **outside the try/catch** with no chunking (`inputs: newRecordIds.map(...)`). HubSpot's batch/read caps at 100 inputs; an import adding 500 contacts between polls (or the first-poll flood from the `isFirstRun` bug) gets a 400. The throw is uncaught, `sendArray` never runs, the seen-set already excludes those contacts forever, and the final `context.setTimeout` never executes — the trigger is dead until manual flow restart.

Fix: chunk inputs by 100, save state only after a successful emit, and keep the reschedule in a `finally`.

src/appmixer/hubspot/crm/FormSubmission/FormSubmission.js
```diff
@@ -0,0 +1,95 @@
...
+            } while (after && !done);
+
+        } catch (err) {
+            await context.log({ step: 'hubspot-form-submission-poll-error', formId, error: err.message });
+            // Reschedule even on error.
+            return context.setTimeout({}, context.config?.pollIntervalMs || DEFAULT_POLL_INTERVAL_MS);
+        }
+
+        // Update state with new last seen timestamp.
+        await context.saveState({ lastSeenAt: newLastSeenAt, isFirstRun: false });
```
apx-vero: **🟠 One transient error after the catch block permanently kills the trigger.**

The try/catch only wraps the pagination loop. This `saveState` and the `sendArray` below run unprotected, *before* the rescheduling `setTimeout` — a state-store hiccup or a 429/500 here exits `receive()` without rescheduling, and no future timeout message ever arrives. The trigger is silently dead until the user restarts the flow. Same pattern in `NewContactInList` (saveState + batch/read + sendArray all in the unguarded tail).

The repo's `"tick": true` convention (e.g. `freshdesk/tickets/NewTicket`) lets the engine own the schedule so one failed poll self-heals — consider it for both polling triggers; at minimum move the reschedule into a `finally`.

src/appmixer/hubspot/crm/ContactPropertyChanged/ContactPropertyChanged.js
```diff
@@ -0,0 +1,83 @@
...
+                maxRetryCount: 3
+            });
+
+            for (const [contactId, event] of Object.entries(eventsByObjectId)) {
+                // Only process events for the configured property.
+                if (event.propertyName && event.propertyName !== watchedProperty) {
+                    continue;
+                }
+                const cacheKey = `hubspot-contact-prop-changed-${watchedProperty}-${contactId}`;
+                const cached = await context.staticCache.get(cacheKey);
```
apx-vero, line 51: **🟠 Dedupe cache key isn't scoped per component — two flows watching the same property steal each other's events.**

`staticCache` is shared across component instances, and this key contains no `componentId`/`flowId`. Flows A and B both watching `email`: `triggerListeners` delivers the same payload to both; A processes first and sets the cache entry; B then sees `cached && event.occurredAt <= cached` and skips the contact entirely. Which flow fires is a race. The `context.lock(context.componentId)` is per-component, so it doesn't serialize across instances.

(`UpdatedContact` has the same unscoped-key pattern, so this is a copied pre-existing bug — but worth fixing in the new code: include `context.componentId` in the key.)

src/appmixer/hubspot/crm/ContactPropertyChanged/ContactPropertyChanged.js
```diff
@@ -0,0 +1,83 @@
...
+            propertiesToReturn = properties.split(',');
+        }
+
+        const { data } = await this.hubspot.call('post', 'crm/v3/objects/contacts/batch/read', {
+            inputs: matchingContactIds.map((id) => ({ id })),
+            properties: propertiesToReturn
+        });
+
+        await context.sendArray(data.results, 'contact');
+
```
apx-vero, line 89: **🟠 Missing creation-event filter — newly created contacts fire as "property changed".**

This `receive()` is a near-verbatim copy of `UpdatedContact.receive()`, but drops its creation filter:

```js
// UpdatedContact only:
if (contact.updatedAt !== contact.createdAt) { results.push(contact); }
```

HubSpot emits `contact.propertyChange` events for initial property values on contact creation, so every contact created with the watched property set will spuriously trigger this component. The 4.5.0 changelog records this exact dedup-vs-creation issue being fixed in `UpdatedContact` — the copy already diverged from its source, which is the argument for extracting the shared lock/dedupe/batch-read pipeline into a helper instead of a third copy.

src/appmixer/hubspot/crm/NewContactInList/NewContactInList.js
```diff
@@ -0,0 +1,96 @@
...
+            propertiesToReturn = properties.split(',');
+        }
+
+        // Fetch all current list members via join-order endpoint (newest first).
+        // We page through and collect IDs not yet seen.
+        let newRecordIds = [];
+        let after = undefined;
+
+        try {
+            do {
```
apx-vero, line 74: **🟠 Entire list re-paginated every poll + unbounded `seenRecordIds` state.**

This loop has no early exit even though join-order is sorted by membership time (the comment above even says "newest first"), and `seenRecordIds` is never pruned — every member ID ever seen is serialized through `saveState` on every poll.

For a 100k-member list: ~1,000 sequential GETs per 5-min poll — at the connector's quota of 5 req/s (`quota.js`) that's ~200 s of quota saturation starving every other HubSpot component, ~288k calls/day from one trigger — plus a 100k-element array written to state every poll, growing forever.

Since the endpoint is ordered by join time, a join-timestamp watermark with early break (the pattern `FormSubmission` already uses with `lastSeenAt`) makes this O(new members) per poll and O(1) state.

apx-vero (commented):
## Automated code review (high effort)

7 finder angles + per-finding verification against the checked-out PR head. **All findings below were CONFIRMED by reading the actual code paths** — none are speculative. The three most severe ones mean each of the three new triggers fails in its primary scenario.

### Finding that can't be anchored inline (unchanged line)

**`routes.js:102` — `_.keyBy(filteredEvents, 'objectId')` destroys per-property events that `ContactPropertyChanged` depends on.** HubSpot sends one `contact.propertyChange` event *per changed property*; `keyBy` keeps only the **last** event per objectId in a webhook batch. When a user edits a contact changing several properties in one save (the common case), the surviving event is arbitrary — whenever it isn't the watched property's event, `ContactPropertyChanged.receive` hits `event.propertyName !== watchedProperty` and silently skips the contact. This collapse was harmless for `UpdatedContact` (any property counts) but is lossy for the new per-property trigger. The fan-out needs to carry an **array of events per objectId**.

### Lower-severity confirmed findings (no inline anchor)

- **Orphaned subscriptions accumulate forever**: every distinct custom property ever watched creates a permanent enabled app-wide HubSpot subscription. `stop()` (inherited) removes only the Appmixer listener; there is no `onListenerRemoved`, no refcount, no cleanup. Pre-existing for the fixed default set, but this PR extends it to an unbounded user-supplied set — pushing toward HubSpot's per-app subscription cap and growing webhook volume for all tenants permanently.
- **`NewContactInList` calls `getObjectProperties` on every poll** even with zero new records. The commons cache TTL is 60 s vs a 5-min poll interval, so it's a real `GET /crm/v3/properties/contacts` every poll. Move it inside `if (newRecordIds.length)`.
- **Reuse**: both hand-rolled `do/while` pagination loops duplicate the existing `Hubspot.paginatedCall`; `ContactPropertyChanged.start()/receive()` fork `BaseSubscriptionComponent.start()` and `UpdatedContact.receive()` almost verbatim (see inline comment about the divergence this already caused). A `getListenerParams()` hook on the base class + a shared dedupe/batch-read helper would shrink the new component to the ~10 lines that actually differ. `DEFAULT_POLL_INTERVAL_MS` is declared twice — belongs in `commons.js`.

Overall: findings on the subscription reconciliation + `keyBy` mean the custom-property feature doesn't work end-to-end even on the happy path; the first-run bugs make the two polling triggers actively harmful or lossy. Recommend rework before merge.

src/appmixer/hubspot/routes.js
vtalas: Fixed in 6a94c3a7: reconciliation now diffs on the full (eventType, propertyName) pair (subKey map), and the disabled-sub path re-enables every matching property without returning early. Verified live — ContactPropertyChanged E2E flow passes on a portal with pre-existing propertyChange subscriptions.

src/appmixer/hubspot/routes.js
vtalas: Fixed in 6a94c3a7: the listener now waits for the lock (retryDelay 500ms, maxRetryCount 20) instead of swallowing 'locked', and rethrows on failure so registration errors surface.

src/appmixer/hubspot/crm/NewContactInList/NewContactInList.js
vtalas: Fixed in 6a94c3a7: seenRecordIds/isFirstRun replaced by a lastJoinedAt watermark seeded in start(), so the first poll emits nothing and state is O(1).

src/appmixer/hubspot/crm/FormSubmission/FormSubmission.js
vtalas: Fixed in 6a94c3a7: isFirstRun removed entirely — lastSeenAt seeded to now() in start() is the only baseline, so submissions between start and first poll are emitted.

src/appmixer/hubspot/crm/NewContactInList/NewContactInList.js
vtalas: Fixed in 6a94c3a7: batch/read is chunked by 100 (readContactsBatch), the watermark is saved only after a successful emit, and the reschedule lives in finally.

src/appmixer/hubspot/crm/FormSubmission/FormSubmission.js
vtalas: Fixed in 6a94c3a7: the whole poll body is inside try/catch and the reschedule setTimeout is in finally, so a transient failure self-heals on the next tick.

src/appmixer/hubspot/crm/ContactPropertyChanged/ContactPropertyChanged.js
vtalas: Fixed in 6a94c3a7: the dedupe cache key includes context.componentId, so two flows watching the same property no longer consume each other's events.

src/appmixer/hubspot/crm/ContactPropertyChanged/ContactPropertyChanged.js
vtalas: Fixed in 6a94c3a7: receive() filters contact.updatedAt !== contact.createdAt (same guard as UpdatedContact), so creation events don't fire the trigger.

src/appmixer/hubspot/crm/NewContactInList/NewContactInList.js
vtalas: Fixed in 6a94c3a7: the poll uses the join-order watermark with early break (O(new members) per poll) and no longer stores a growing seen-set.

src/appmixer/hubspot/engagements/CreateEmail/component.json
```diff
@@ -221,6 +221,10 @@
                                 "label": "Timestamp",
                                 "value": "hs_timestamp"
                             }
+                        },
+                        "example": {
+                            "hs_note_body": "Follow-up scheduled",
+                            "hs_lastmodifieddate": "2026-07-24T12:00:00.000Z"
                         }
```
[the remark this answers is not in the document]
vtalas: Fixed in e5d4d245: CreateEmail properties example now uses email engagement fields (hs_email_direction/subject/status, hs_timestamp).

src/appmixer/hubspot/crm/FormSubmission/component.json
```diff
@@ -0,0 +1,116 @@
...
+                        "example": "a2c3e5f7-1b2d-4e6f-8a9b-0c1d2e3f4a5b"
+                    }
+                },
+                {
+                    "label": "Submitted At",
+                    "value": "submittedAt",
+                    "schema": {
+                        "type": "string",
+                        "example": 1784548800000
+                    }
```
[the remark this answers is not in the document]
vtalas: Fixed in e5d4d245: submittedAt declared as number (epoch ms) with a numeric example and relabeled 'Submitted At (epoch ms)'.

vtalas (approved):

PR #1125, opened 2026-06-04
Microsoft Teams SendChannelMessage: support HTML content type for hyperlinks

petrolivka:
## Summary

Adds an optional **Content Type** (Text/HTML) option to the Microsoft Teams `SendChannelMessage` component so users can send rich text and **clickable hyperlinks**.

Previously the component sent the Microsoft Graph request body without `body.contentType`. Graph defaults that to `text`, so HTML such as `<a href="...">Text</a>` rendered as literal plain text instead of a clickable link.

## Changes

- **`SendChannelMessage.js`** — new optional `contentType` param (default `'text'`), passed through to the Graph request as `body.contentType`. Backward-compatible: existing flows keep plain-text behavior.
- **`component.json`** — new `Content Type` select (Text/HTML, default Text) + schema enum `["text", "html"]`. Tooltip on `Content` updated with a hyperlink example.
- **`bundle.json`** — version bump `1.2.4` → `1.2.5` + changelog entry.

## Notes

- Markdown (`[Text](URL)`) is **not** supported by the Graph channel-messages endpoint — only `text` or `html`.
- Adaptive Cards use the `attachments` array (out of scope here).

## Docs

Microsoft Graph — [channel: post messages](https://learn.microsoft.com/en-us/graph/api/channel-post-messages?view=graph-rest-1.0&tabs=http)

## Testing

- JSON of changed files validated (parses OK).
- Lint clean.

Reported via Freshdesk #9647.
Closes Appmixer-ai/appmixer-components#2724


petrolivka:
Superseded by #1327 (phase B of Appmixer-ai/appmixer-components#2858), which adds this `Content Type` input together with `Reply To Message ID`, `Subject` and `Importance`, rewritten on top of the shared `makeRequest` client that phase A (#1326) introduced.

The behaviour you added here is preserved: `contentType` defaults to `text`, HTML is offered for rich text and clickable links, and the tooltip still notes that Microsoft Teams does not support Markdown.

Closing in favour of #1327 — thanks, this was the right gap to spot.

PR #1149, opened 2026-06-20
zoho books: new invoice

apx-vero:


apx-vero:
Superseded by #1336, which carries this work together with the rest of the open Zoho changes.

PR #1150, opened 2026-06-20
google.calendar: add AcceptEvent action, UpdatedEvent & DeletedEvent triggers

apx-vero:
## Summary

Implements the Google Calendar components requested in Appmixer-ai/appmixer-components#2741.

The issue asked for four items; **Decline Event** already exists in the connector (`DeclineEvent`), so this PR adds the remaining three:

| Component | Type | Description |
|-----------|------|-------------|
| **AcceptEvent** | Action | Accepts a calendar event invitation by patching the attendee's `responseStatus` to `accepted`. Mirrors the existing `DeclineEvent` action (same inputs: Calendar, Event, Attendee Email, Send Updates). |
| **UpdatedEvent** | Trigger | Polling trigger that fires when an existing event changes (modified after the baseline and created on/before it, so it doesn't overlap with `NewEvent`). |
| **DeletedEvent** | Trigger | Polling trigger that fires when an event is deleted (reported by the Google Calendar API as a `cancelled` event via `showDeleted: true`). |

## Implementation notes

- `AcceptEvent` validates required inputs with `context.CancelError` and returns the patched event.
- Both new triggers use `tick()` with `loadState()`/`saveState()` to track a `since` baseline and avoid duplicates, following the existing `NewEvent` pattern.
- Both triggers include a `test()` method for Flow Test Mode (read-only fetch of the newest matching event), consistent with the 1.4.0 work.
- Bundle bumped to `1.5.0` with changelog entries.

## Validation

- `npm run lint` — passes

Source issue: https://github.com/Appmixer-ai/appmixer-components/issues/2741


src/appmixer/google/calendar/AcceptEvent/component.json
```diff
@@ -0,0 +1,242 @@
...
+                        "source": {
+                            "url": "/component/appmixer/google/calendar/ListCalendars?outPort=out",
+                            "data": {
+                                "transform": "./transformers#calendarsToSelectArray"
+                            }
+                        },
+                        "tooltip": "The calendar ID."
+                    },
+                    "eventId": {
+                        "type": "select",
```
vtalas: @apx-vero  This shouldn't be "select"; it should be text. Otherwise, the user won't have a chance to insert value manually.

vtalas (approved):

PR #1205, opened 2026-08-10
deepgram(minor): add Deepgram voice-AI connector (STT, TTS, Text Intelligence)

apx-vero:
## Summary

Adds a new **Deepgram** connector (`src/appmixer/deepgram`) — a voice-AI platform for speech-to-text (Nova-3), text-to-speech (Aura-2) and text intelligence (summarization, sentiment, topics, intents).

Implements the **Phase 1 MVP** plus the **recommended polling triggers** from the issue. Phase 2 (project/key/member/usage/billing management) and Phase 3 (Voice Agent config) are intentionally deferred per the issue's recommendations.

## Authentication

- **API key only** (no OAuth2). Key is sent as `Authorization: Token <key>` — the `Token` scheme, **not** `Bearer`.
- Validation uses the cheap, side-effect-free `GET /v1/projects`.
- **Region selector** on the auth account: Global / EU / AU / Custom (dedicated or self-hosted host). Deepgram keys are portable across regions, so region only changes the base URL (`lib.getBaseUrl`).

### Actions (`core/`)
| Component | Endpoint | Notes |
|---|---|---|
| **TranscribeAudio** | `POST /v1/listen` | Synchronous. Accepts an audio URL (JSON) **or** a file streamed from FileStorage. Exposes model/language/format/diarize + Audio Intelligence flags. 504 errors point the user at the async component. |
| **TranscribeAudioAsync** | `POST /v1/listen?callback=…` | Returns `request_id` immediately; requires a callback URL (documents the 80/443/8080/8443 port restriction). |
| **TextToSpeech** | `POST /v1/speak` | Returns **binary audio** → saved via `context.saveFileStream()`, outputs a `fileId`. Validates the 2000-char limit client-side. |
| **AnalyzeText** | `POST /v1/read` | Text or URL; summarize/sentiment/topics/intents. |
| **ListModels** | `GET /v1/models` | `outputType`; powers the STT model / TTS voice typeaheads. |
| **GetModel** | `GET /v1/models/{id}` | |
| **ListProjects** | `GET /v1/projects` | `outputType`; source for the project typeahead. |
| **GetProject** | `GET /v1/projects/{id}` | |
| **FindRequests** | `GET /v1/projects/{id}/requests` | `outputType` + `notFound`; date range + status filter. |
| **GetRequest** | `.../requests/{id}` | Collects the stored result of an async transcription. |
| **MakeApiCall** | any | Standard passthrough (region-aware base URL). |

### Triggers (polling)
- **NewRequest**, **TranscriptionCompleted** (optional full-result fetch), **FailedRequest** — all `tick()`-based, deduping by `request_id` (sort order of `/requests` is undocumented, so no reliance on ordering/cursors). Each has a Flow Test Mode `test()` method sharing the production fetch path.

## Design notes / issue findings addressed
- **Sync 10-min ceiling** — the 504 message explicitly names `Transcribe Audio (Async)`.
- **Error normalization** in `lib.js` tolerates both Deepgram error envelopes (`err_code/err_msg` and `category/message/details`) and non-JSON bodies, with actionable messages for 401/402/413/429/504.
- **Project-scoped management** — `projectId` is a typeahead (text + dynamic source) so users can pick from their projects or enter an ID manually (per the `no-select-with-source` standard).
- **Webhooks** cannot self-register in Deepgram, so v1 ships polling triggers only (webhook trigger deferred to phase 2, as recommended).

## Validation
- `node scripts/validate.js --connector deepgram` → 0 failures, 0 warnings
- `node scripts/validate.js --changed --base origin/dev` → passed
- `npm run lint` → clean

## Deliberately out of scope (per issue)
Phase 2 management components, Phase 3 Voice Agent config, `/v1/auth/grant`, and all WebSocket surfaces (live STT/TTS, Flux, Voice Agent conversation).

## Note on untrusted issue content
The source issue body was treated strictly as a task description. It contained no instructions attempting to exfiltrate secrets, change git/CI config, or contact external hosts, so nothing needed to be ignored on that front.

Closes Appmixer-ai/appmixer-components#2785


apx-vero:
## Pre-release pass — two commits added

`node scripts/validate.js --connector deepgram` 0/0, eslint clean, and **all 9 E2E flows pass** on `dev-automated-00001` against the published module, including the Transcribe Audio submit/callback round trip and both polling triggers.

### 4430f910 — vacuous asserts, stale references

Both trigger E2E flows carried regex assertions keyed on `"value"` while the Assert component reads `"regex"`. As written they ran `new RegExp(undefined)`, matched anything and **passed vacuously** — each flow was effectively asserting only its `notEmpty` half. Fixed (3 occurrences), and they still pass with real assertions.

Also: four references to `Transcribe Audio (Async)`, a component that no longer exists since 440da1b4 collapsed the sync/async pair; `AfterAll`/`ProcessE2EResults` sitting 80 px from both assert rows (layout rule wants same-row or ≥ 128); and the leftover `artifacts/icon.jpeg` from the switch to SVG.

### 09fcd837 — Transcribe Audio under concurrent load

The component was a faithful implementation of the documented self-callback pattern. Auditing it for load found five defects, four of them inherent to the pattern's use of component state as the echo carrier:

1. **The echo now rides in the callback URL** (`?echo=<json>`, read back from `context.messages.webhook.content.query`) — no component state at all. Deepgram starts processing the moment it accepts the job, so the callback raced the `stateSet` that followed the submit. Measured margin on the 26 s E2E clip: **~0.5 s** (`out`→`done` 538–673 ms across five runs); a one-second clip plus a loaded state store closes it. Losing that race dropped exactly the field the mechanism exists to deliver, **and leaked the entry permanently** because the callback's `stateUnset` ran before the submit's `stateSet`. State has no TTL either, so a job that never called back leaked forever.
2. **A callback with no `metadata.request_id` is logged and acknowledged**, not emitted. Anything can POST to a webhook URL, and downstream could not tell a stray request's empty transcript from a real one.
3. **`context.response()` moved into a `finally`** — a throw between the emit and the ack (`stateUnset` was one) meant no 2xx, so Deepgram redelivered and the redelivery re-ran the failure.
4. **The upload stream is destroyed when the submit fails.** On a 413/429/5xx it was left open holding a descriptor, and an auto-retried component opened a new one per attempt.
5. **A submit returning no `request_id` now throws** instead of sending `request_id: undefined` to `out` and silently unlinking the callback.

Plus `quota.js`: the `limit-concurrency` comment claimed it capped in-flight work. True while the component blocked on the synchronous endpoint; it now returns in ~2 s, so the slot is released while the job keeps running on Deepgram's side and the rule caps concurrent *submissions* only. Comment rewritten to say what it does — the sliding window is what bounds the rate.

**What proves the URL-carried echo works:** the Transcribe Audio flow's `equal` assertion on the echoed `correlationId` over the `done` port. It passes, so the query string survives the round trip through Deepgram.

The pattern reference this connector was built from is fixed too: Appmixer-ai/appmixer-skills#15.

### Two things for the reviewer, not fixed here

- **`appmixer e2e import` needs `--no-validate`** for this connector. The CLI rule `dynamic-source-static-input` flags `FindRequests.projectId`, but that component's outPort source pins `"in/projectId": "dummy"` and never reads the inspector — the server-side variable check (which runs regardless) passes clean. ~45 connectors use the same literal-pin trick, so the rule needs teaching in appmixer-cli (`sourceMappedFields()` should only collect fields whose mapped value starts with `inputs/`). Hardcoding a tenant project id in the flow would trade a false positive for a real portability bug.
- **`MakeApiCall` bypasses `lib.normalizeError`.** The same 415 reads `Deepgram API request failed (415). remote server failed to offer audio data...` through every other component but `Request failed with status code 415` through MakeApiCall, and as a plain retryable `Error` rather than a `CancelError`. Left alone because it is a design call about how raw a raw-API-call component should be — happy to change it if you want consistency.

### Failed Request trigger has no E2E flow, and that is correct

I verified the premise behind dropping it in 091cc803: Deepgram rejects a bad audio URL **synchronously** (415, `remote server failed to offer audio data`), so no job is ever created and nothing lands in the failed-request log — the project has zero failed entries historically. The trigger cannot be provoked from the connector's own components.

vtalas (approved):

PR #1207, opened 2026-08-11
hubbi(minor): add HubBI Flows API connector (start hubs, push records, hub event trigger)

apx-vero:
## HubBI — Flows API connector

Adds `src/appmixer/hubbi`, a new **action + trigger** connector for HubBI's Flows API — the surface HubBI exposes for driving its "hubs" (data-conversion pipelines) from the outside. The API is small and purpose-built (no CRUD, no read API for past events, no pagination), so the connector maps onto it one-to-one.

Implements the connector described in Appmixer-ai/appmixer-components#2787.

### Authentication (`apiKey`)
HubBI is deployed per tenant, so the base URL is part of the credentials:

| Field | Notes |
|---|---|
| `baseUrl` | Tenant base URL, e.g. `https://test-app.hubbi.nl` |
| `token` | HubBI JWT, sent as `Authorization: Bearer <token>` |
| `clientKey` | Client identifier (UUID), sent as the `clientKey` query param on every call |

`validate`/`requestProfileInfo` probe `ListTargetHubs`; the account name masks the client key to first/last three characters (`HubBI (abc...123)`). No OAuth2 — HubBI exposes no authorization-code flow for the Flows API.

### Components

| Component | Type | Public | Purpose |
|---|---|:-:|---|
| `StartHub` | Action | ✅ | Start a hub with no payload (`HubsStart`) |
| `StartHubWithData` | Action | ✅ | Start a hub with one or more records (`HubsStartWithData`, POST) |
| `NewHubEvent` | Trigger (webhook) | ✅ | Fires when HubBI posts a hub event |
| `MakeApiCall` | Action | ✅ | Arbitrary authenticated Flows API request |
| `ListSourceHubsWithPostData` | List | — | Hub picker source |
| `ListSourceHubsWithoutPostData` | List | — | Hub picker source |
| `ListTargetHubs` | List | — | Hub picker source + auth validation |
| `GetSourceFields` | List | — | Source field definitions |
| `GetTargetFields` | List | — | Target field definitions |

### Design notes
- **Input/output shapes are generated from the hub, not hardcoded.** `StartHubWithData` builds its inspector from the selected hub's `SourceFields` (with a raw-JSON toggle for bulk pushes); `NewHubEvent` builds its output port options from `TargetFields`.
- **.NET type mapping.** `lib.mapFieldType()` maps HubBI's .NET type names (`Int32`, `Double`, `DateTime`, `Guid`, `Boolean`, …) to an Appmixer inspector type and a JSON Schema type; unknown types degrade to `string`.
- **Bulk records.** `StartHubWithData` sends a JSON **array** body in one request and reports the record count.
- **Custom retry classification.** `lib.rethrowHubbiError()` rethrows **409 Conflict** as a plain `Error` (engine treats as unknown → retries) and converts **423 Locked** to `CancelError` (engine does not retry).
- **Manual webhook registration.** HubBI has no API to register webhooks, so `NewHubEvent.start()` only records the URL in state and the trigger surfaces a read-only **Webhook URL** for the user to paste into HubBI.
- **Foreign/empty events acknowledged, not emitted.** Events for a different `conversionKey`, or with no records, are logged and answered with `context.response()`.
- **`test()` synthesizes rather than fetches.** No endpoint exists for past events, so Flow Test Mode builds a sample record from the hub's target field definitions.
- **Quota:** 5 req/s, FIFO (`quota.js`).

### Follow-ups from the issue addressed here
- Hub pickers are `text` (typeahead) bound to dynamic sources, so a conversion key can still be typed by hand (fixes `no-select-with-source`).
- `GetSourceFields`/`GetTargetFields` pass their required `conversionKey` to the port-options request (fixes `dynamic-outport-required-inputs` for those).
- `MakeApiCall` added (satisfies `connector-has-makeapicall`).
- `version` set in `service.json`.

### Validation
- `npm run lint` — clean.
- `node scripts/validate.js --connector hubbi` — **1 remaining advisory**: `dynamic-outport-required-inputs` recommends `ignoreAuth=true` on `NewHubEvent`'s output-port source. This is left **intentionally authenticated**: unlike the List helpers (whose options come from a static schema), `NewHubEvent`'s option generation genuinely calls `TargetFields` and needs `context.auth`, exactly as the issue's §7.2 caveat warns. The call is guarded so it degrades to a generic option if fields can't be fetched.

### Notes / assumptions for review
- The exact JSON shapes of the HubBI list/field endpoints were not verifiable without a live tenant; the connector normalizes both PascalCase and camelCase defensively (`lib.normalizeHubs`/`normalizeFields`). Worth confirming against a real tenant.
- Bundle version is `2.0.0` per the issue (the connector carried 1.x versions during earlier development, including the removed `HubsList` endpoint).
- No E2E flows and no live-tenant verification in this PR — those are handled in a separate step and require a HubBI test tenant.

Closes Appmixer-ai/appmixer-components#2787

### DIFF
--- a/src/appmixer/hubbi/artifacts/icon.svg
+++ b/src/appmixer/hubbi/artifacts/icon.svg
@@ -1,0 +1,14 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 64 64" width="64" height="64">
+  <rect width="64" height="64" rx="12" fill="#0B5FFF"/>
+  <circle cx="32" cy="32" r="7" fill="#FFFFFF"/>
+  <circle cx="14" cy="16" r="5" fill="#FFFFFF"/>
+  <circle cx="50" cy="16" r="5" fill="#FFFFFF"/>
+  <circle cx="14" cy="48" r="5" fill="#FFFFFF"/>
+  <circle cx="50" cy="48" r="5" fill="#FFFFFF"/>
+  <g stroke="#FFFFFF" stroke-width="3" stroke-linecap="round">
+    <line x1="27" y1="27" x2="18" y2="19"/>
+    <line x1="37" y1="27" x2="46" y2="19"/>
+    <line x1="27" y1="37" x2="18" y2="45"/>
+    <line x1="37" y1="37" x2="46" y2="45"/>
+  </g>
+</svg>
--- a/src/appmixer/hubbi/artifacts/test-flows/test-flow-makeapicall.json
+++ b/src/appmixer/hubbi/artifacts/test-flows/test-flow-makeapicall.json
@@ -1,0 +1,178 @@
+{
+    "name": "E2E HubBI - MakeApiCall list target hubs",
+    "type": "automation",
+    "flow": {
+        "7a205352-f70f-42a5-9f4a-4a4ea6649af2": {
+            "type": "appmixer.utils.controls.OnStart",
+            "x": 64,
+            "y": 16,
+            "source": {},
+            "version": "1.0.0",
+            "errorHandling": {
+                "autoRetry": false,
+                "onError": "stopFlow"
+            }
+        },
+        "c2e7535e-76ae-4189-b362-b51eae89ca84": {
+            "type": "appmixer.hubbi.core.MakeApiCall",
+            "x": 320,
+            "y": 16,
+            "source": {
+                "in": {
+                    "7a205352-f70f-42a5-9f4a-4a4ea6649af2": [
+                        "out"
+                    ]
+                }
+            },
+            "version": "2.0.0",
+            "config": {
+                "transform": {
+                    "in": {
+                        "7a205352-f70f-42a5-9f4a-4a4ea6649af2": {
+                            "out": {
+                                "type": "json2new",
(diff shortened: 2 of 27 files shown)


apx-vero:
E2E report: Appmixer-ai/appmixer-components#2787 (comment https://github.com/Appmixer-ai/appmixer-components/issues/2787#issuecomment-5454521104).

**3/3 flows now pass.** All three failed as delivered, for three separate reasons — `ProcessE2EResults` filed results under `{{{tc-flowname}}}` instead of the flow name; the asserts carried an off-spec `type` field that can never match a mustache-rendered value (`expected 'true' to be a boolean`); and both StartHub flows used placeholder conversion keys. Fixed in `7215bb48` (pushed to this branch).

Worth noting from that third one: `POST /Flows/Home/StartHub` returned `{"started": true}` for a conversion key that does not exist, so that flow was passing while proving nothing. Both flows now use real hubs from the E2E account.

Three orphan flows from an earlier iteration of this PR are still on the instance with stale green results — commands to remove them are in the issue comment.

vtalas (approved):

PR #1211, opened 2026-08-13
microsoft: surface Graph error body in makeRequest instead of bare status code

apx-vero:
## Summary

`makeRequest` in the Microsoft connector modules wrapped `context.httpRequest` (a bare axios instance) without a try/catch, so a non-2xx Graph response propagated as a raw `AxiosError` whose message is only `Request failed with status code <n>`. Everything needed to diagnose the failure — the Graph `error.code` (e.g. `MailboxNotEnabledForRESTAPI`), the human-readable `error.message`, and the `request-id` header Microsoft support asks for — was dropped. This is what left the Freshdesk ticket 9719 customer unable to tell why their `SendEmail` (POST /me/sendMail) call 404'd.

## Change

- Added a shared `graphError(error)` helper to `microsoft/microsoft-commons.js` that re-throws an `Error` carrying the Graph `error.code`, `error.message` and the `request-id` response header. When the response has no parseable Graph error body (non-JSON body, network error, etc.) the **original error is returned unchanged**, so nothing is swallowed and no `undefined` leaks into the message.
- Wrapped the `makeRequest` calls in `mail`, `calendar` and `teams` `commons.js` with it.

Example resulting message:

```
Microsoft Graph error MailboxNotEnabledForRESTAPI: The mailbox is either inactive, soft-deleted, or is hosted on-premise. (request-id: 8f2c...)
```

## Scope notes

The issue asked to check `dynamics`, `sharepoint`, `onedrive`, `teams`, `calendar` and any other `microsoft/*/commons.js` for the same unwrapped pattern. After review:

- **mail, calendar, teams** — each has the identical unwrapped `makeRequest` wrapping `context.httpRequest`. Fixed.
- **onedrive / sharepoint** — do not use this `makeRequest` pattern; they go through the `onedrive-api` library wrapped by the existing `commons.formatError` helper, which already extracts the API error message. Left as-is.
- **dynamics** — has no `makeRequest` wrapper (calls `context.httpRequest` directly in scattered places); out of scope for this focused fix.

The related `getMailAttachments` base64 size-guard / upload-session gap called out under "out of scope" in the issue is **not** addressed here and should be tracked separately.

## Validation

- `npm run lint` — passes
- Patch-bumped `bundle.json` for mail (2.0.1), calendar (1.2.2) and teams (1.3.1) with changelog entries.

Closes Appmixer-ai/appmixer-components#2789

### DIFF
--- a/src/appmixer/microsoft/artifacts/test-flows/test-flow-calendar-event-lifecycle.json
+++ b/src/appmixer/microsoft/artifacts/test-flows/test-flow-calendar-event-lifecycle.json
@@ -251,10 +251,12 @@
                             "out": {
                                 "type": "json2new",
                                 "modifiers": {
-                                    "outputType": {}
+                                    "outputType": {},
+                                    "maxRecords": {}
                                 },
                                 "lambda": {
-                                    "outputType": "array"
+                                    "outputType": "array",
+                                    "maxRecords": 5
                                 }
                             }
                         }
@@ -491,7 +493,7 @@
                                                     "name": "g_jsonPath",
                                                     "params": [
                                                         {
-                                                            "value": "$[0].id"
+                                                            "value": "$.result[0].id"
                                                         }
                                                     ]
                                                 }
--- a/src/appmixer/microsoft/artifacts/test-flows/test-flow-calendar-event-lifecycle.json
+++ b/src/appmixer/microsoft/artifacts/test-flows/test-flow-calendar-event-lifecycle.json
@@ -1,0 +1,637 @@
+{
+    "name": "E2E Microsoft Calendar - Event Lifecycle",
+    "type": "automation",
+    "flow": {
+        "3a43a949-d14e-4354-86c9-100e3ac8350c": {
+            "type": "appmixer.utils.controls.OnStart",
+            "x": 100,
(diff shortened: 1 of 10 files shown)


apx-vero:
E2E report: Appmixer-ai/appmixer-components#2789 (comment https://github.com/Appmixer-ai/appmixer-components/issues/2789#issuecomment-5454526274).

**Calendar and mail pass; teams is blocked on the account, not on this code.** All three flows failed as delivered — every List assert read `$[0].id` while `ListEvents`/`ListEmails`/`ListChannelMessages` emit `{result}`/`{emails}`/`{messages}` respectively, and `ListEmails` at its default limit of 100 overflowed the engine's 2 MB message cap at 9.97 MB. Fixed in `a0c061dc` (pushed to this branch).

Teams cannot start: neither `appmixer:microsoft` account on the instance has Teams consent (`Team.ReadBasic.All`, `Channel.ReadBasic.All`, `ChannelMessage.Send`), so `startFlow` fails with `Access token not found for component … ListTeams`. Needs a re-consent in the designer, then the flow runs as is.

vtalas (approved):

PR #1213, opened 2026-08-18
strava(patch): centralize API base URL, fix FindActivityStreams and dynamic output pickers

apx-vero:
Closes the API base URL work from #2792, and fixes the defects that surfaced while testing it end to end.

## API base URL

Strava is migrating from `https://www.strava.com/api/v3` to `https://api-v3.strava.com`, available January 4, 2027. The old host stays the only working one until then, so this centralizes it into `constants.js` (`API_BASE_URL`) and routes all nine components plus `auth.js` through it. The migration then becomes a one-line change. No behavior change now.

`oauth/authorize` deliberately stays on `www.strava.com` — it is not under `/api/v3`.

## Fixes found while testing

**FindActivityStreams was broken on both paths.**

- Strava answers **404** for an activity that carries no stream data — which is *every* manually created activity. The component let that throw and failed the whole flow, instead of routing to the `notFound` port it already declares.
- The request asked for `key_by_type=true` (an object keyed by stream type) while the component declared the array-shaped schema, length-checked an array and passed it to `sendArrayOutput`. `records.length` on an object is `undefined`, so `records[0]` was `undefined` and the `out` port emitted an empty record. The success path could not have worked.

**Dynamic output port sources could not run** for `ListActivities`, `FindActivities` and `FindActivityStreams`. They required an auth account even though schema generation returns static metadata and returns before any API call, and `FindActivityStreams` never passed its required inputs into the source call. Either alone made the call fail, so the designer's variable picker degraded to Raw Output.

Verified against the live instance, with no account bound:

| | before | after |
|---|---|---|
| no `ignoreAuth` | `401 Access token not found` | — |
| `ignoreAuth`, no required inputs | `500 Validation error on ports: in` | — |
| both fixed | — | **200**, 11 / 15 / 7 fields |

(`keys` is an array, so its placeholder is `[]` — a string literal fails the engine's port validation.)

## Test flows

Four E2E flows covering all nine components. Two defects in them:

- **Three `regex` assertions used the key `expected` while the Assert component reads `regex`** — they ran as `new RegExp(undefined)`, matched anything and passed vacuously.
- Read Activities now asserts **both** outcomes of `FindActivityStreams`, so it holds whether or not the account has a stream-bearing activity.

Layout normalized to the 208/128 grid.

## Test results

All four pass against a live Strava account:

```
✓ E2E Strava - Read Activities
✓ E2E Strava - Activity Lifecycle
✓ E2E Strava - Athlete Profile and Stats
✓ E2E Strava - Make API Call
```

`validate.js --connector strava` and `eslint` are clean.

## Notes for the reviewer

- **Needs appmixer-cli#444.** The `dynamic-source-static-input` rule failed on `FindActivityStreams.activityId`: it read only the *keys* of `source.data.messages`, never the value, so it could not tell `"inputs/in/activityId"` (reads the inspector — rule is right) from `"dummy"` (a pinned literal — rule is wrong). Fixed in Appmixer-ai/appmixer-cli#444, which drops 69 false positives across 22 connectors while keeping the 14 genuine ones. Until that lands, `appmixer e2e import` here needs `--no-validate` — a blunt last resort, since it disables all 27 rules and the suite has no ignore-list.
- **The `out` branch of `FindActivityStreams` is still unexercised live.** The test account holds only manually created activities, which never have streams, so the flow legitimately takes `notFound`. Recording or uploading one real GPS activity to that account would exercise the success path too.
- **`Activity Lifecycle` leaves data behind.** The Strava v3 API has no delete-activity endpoint, so the connector has no Delete component and each run leaves one activity on the account. The `create-without-cleanup` warning is unavoidable.
- **Strava API access now requires a paid subscription** on the athlete who owns the API application; without it the app goes *Inactive* and every endpoint 403s while OAuth still succeeds. Documented in `context.md`, since it reads exactly like a connector bug.

https://claude.ai/code/session_013AWZrdBSjYx8VbiUe7p4g8


apx-vero:
Published E2E flow(s) (not started):

- **E2E Strava - Activity Lifecycle** — https://dev-automated-00001.dev.appmixer.ai/designer/1546434e-7192-4231-b5bf-d8a4c93bbe98
- **E2E Strava - Athlete Profile and Stats** — https://dev-automated-00001.dev.appmixer.ai/designer/fd8429d8-af38-4ea2-8af0-c2211078887b
- **E2E Strava - Make API Call** — https://dev-automated-00001.dev.appmixer.ai/designer/026b343b-e04b-42ad-abff-adf63fb598b5
- **E2E Strava - Read Activities** — https://dev-automated-00001.dev.appmixer.ai/designer/4841074a-e74d-4e43-9da6-1d22d8750a6b

vtalas (approved):

PR #1218, opened 2026-08-24
mistral: add Mistral AI connector (appmixer.ai.mistral)

apx-vero:
Adds the **Mistral AI** connector (`appmixer.ai.mistral`), an API-key connector in the flat `ai/` namespace, built to the spec in Appmixer-ai/appmixer-components#2795.

## Components (10)

| Component | Endpoint |
|---|---|
| Send Prompt | `POST /v1/chat/completions` — conversation state via `conversationId` |
| Describe Images | `POST /v1/chat/completions` — vision, base64 data URI |
| Generate Embeddings | `POST /v1/embeddings` |
| Extract Document Text | `POST /v1/ocr` — file *or* URL, page-range parsing |
| Create Transcription | `POST /v1/audio/transcriptions` — Voxtral |
| Create Speech | `POST /v1/audio/speech` — saves the audio as a file |
| Moderate | `POST /v1/moderations` — 11 categories + `flagged` aggregate |
| List Models | `GET /v1/models` |
| List Voices | `GET /v1/audio/voices` |
| Make API Call | canonical #1459 shape |

Text-to-speech is included because it is GA, not beta. Fine-tuning is deliberately **not** implemented — it is deprecated in the Mistral API.

## Model dropdowns are capability-filtered

`/v1/models` returns 56 models with a `capabilities` object. Rather than offering all of them everywhere, each dropdown is filtered to what its endpoint accepts: chat 37, vision 32, OCR 7, transcription 2, speech 2, moderation 1, embedding 4. Embedding models are identified by carrying no capability flag at all — the API marks them by omission.

`ListModels` / `ListVoices` back these dropdowns and cache through `context.staticCache` + `context.lock`, since the designer fires source calls in concurrent bursts.

## Testing

**CLI: 10/10 components pass against the live API.** Seven defects were found and fixed, five of which only a real call could surface:

- **Create Speech was unusable as shipped** — it sent `voice_id` while the documented parameter is `voice` (`"Either ref_audio or voice must be provided"`), had no default model so a text-only request returned 400, and its voice input was free text requiring a UUID no user could guess. It now sends `voice`, defaults to `voxtral-mini-tts-latest`, and offers the `ListVoices` dropdown. Output verified as genuinely playable MP3 (ID3 header), not merely a 200.
- **Moderate declared 9 categories against the API's 11** — `dangerous_and_criminal_content` does not exist (the API splits it into `dangerous` and `criminal`) and `jailbreaking` was missing, so one option was permanently undefined and three real categories unreachable.
- **List Models declared 4 of the 13 fields the API returns**, hiding `capabilities`, `description`, `max_context_length`, `deprecation` and the rest from the variable picker.

**E2E: 4/4 flows pass — 21 assertions, 0 errors**, verified against the stored `ProcessE2EResults` payload rather than the runner's exit code. That mattered: the first run printed `PASSED` on all four while two recorded *zero* assertions, caused by two flow-authoring bugs now fixed (a `regex` clause reads its pattern from `regex`, not `expected`; and `ProcessE2EResults` must report `$.<afterAll>.out`, not the tail of a chain that ends in cleanup).

The vision flow builds its input image from an inline data URI via `utils.converters.DataUri2File`, so it depends on no external host. Both flows that leave a file behind clean it up.

## Note for reviewers

`appmixer e2e import` needs `--no-validate` here. The local rule `static-outport-deep-path` resolves an outPort option by its **first** path segment only, so it rejects the dotted option values this connector declares (`choices.message.content`, `categories.jailbreaking`) — the same convention `ai/openai` and `ai/requesty` use. The server's variable service offers exactly those paths and rejected the shortened form the rule asked for, so the rule is wrong here. Worth fixing in `appmixer-cli` separately.

Lint clean; `node scripts/validate.js --connector ai` reports 0 findings for mistral.

https://claude.ai/code/session_01GpLDcqdduhrUu2TbUfKNeJ


petrolivka:
- https://dev-automated-00001.dev.appmixer.ai/designer/706c1f16-afde-47c2-8224-469d427214f1
- https://dev-automated-00001.dev.appmixer.ai/designer/e80a53e3-bf8e-4f5e-950a-288c3c2312dd
- https://dev-automated-00001.dev.appmixer.ai/designer/690b3308-8818-4bd6-a549-487f6145f8a0
- https://dev-automated-00001.dev.appmixer.ai/designer/a14c2114-9a12-4d62-880e-b0a0779f7280

src/appmixer/ai/mistral/ExtractDocumentText/ExtractDocumentText.js
```diff
@@ -0,0 +1,84 @@
...
+
+        let document;
+        if (file) {
+            const fileInfo = await context.getFileInfo(file);
+            const fileContent = await context.loadFile(file);
+            const base64 = fileContent.toString('base64');
+            let contentType = fileInfo.contentType;
+            if (!contentType) {
+                const ext = path.extname(fileInfo.filename).toLowerCase();
+                contentType = IMAGE_EXTENSIONS.includes(ext) ? `image/${ext.substring(1)}` : 'application/pdf';
```
vtalas: ⚠️ **warning** · `behavior.invalid-mime-fallback`

The contentType fallback builds `image/${ext.substring(1)}`, which yields the invalid MIME type `image/jpg` for a `.jpg` file; that value is not in IMAGE_TYPES, so a JPEG whose stored contentType is missing is sent to the OCR API as `document_url` with a `data:image/jpg;...` URI instead of `image_url`.

**Suggested fix:** Map extensions to real MIME types (`.jpg`/`.jpeg` -> `image/jpeg`) the same way DescribeImages.js:37 already does, rather than deriving the MIME type from the extension string.

src/appmixer/ai/mistral/ListModels/ListModels.js
```diff
@@ -0,0 +1,125 @@
...
+        const { outputType = 'array' } = context.messages.in.content || {};
+
+        // Generate output port options dynamically if requested.
+        if (context.properties && context.properties.generateOutputPortOptions) {
+            return lib.getOutputPortOptions(context, outputType, schema, { label: 'Models' });
+        }
+
+        // https://docs.mistral.ai/api/#tag/models
+        // Cached: this component also backs the model dropdowns of other components.
+        const data = await lib.apiCallCached(context, '/models');
```
vtalas: ⚠️ **warning** · `source.error-suppression-missing`

ListModels and ListVoices back eight inspector dropdowns but neither passes an `isSource` sentinel in `source.data.properties` nor catches fetch errors, so an invalid key, an unsaved auth or a Mistral outage surfaces as an error popup every time a component's inspector is opened.

**Suggested fix:** Pass `"properties": { "isSource": true }` in each dropdown's `source.data` and, when that flag is set, catch the request error and return an empty result instead of throwing — as ai/claude/ListModels/ListModels.js already does.

src/appmixer/ai/mistral/lib.js
```diff
@@ -0,0 +1,120 @@
...
+};
+
+/**
+ * Cached GET for dynamic inspector source calls. The designer fires source calls
+ * in concurrent bursts, so the first caller populates the cache under a lock and
+ * the rest read the cached value.
+ */
+const apiCallCached = async (context, url) => {
+
+    const key = 'ai-mistral-' + url.replace(/\W+/g, '-') + '-' + (context.auth.apiKey || '').slice(-8);
```
vtalas: ⚠️ **warning** · `source.cache-key-collision`

The shared cache key uses only the last 8 characters of the API key, so two accounts whose keys happen to share that suffix would read each other's cached response — which for /audio/voices includes account-private custom voices.

**Suggested fix:** Key the cache on a hash of the full credential, e.g. `crypto.createHash('sha256').update(JSON.stringify({ url, apiKey: context.auth.apiKey })).digest('hex')`, as the reference `callEndpointCached` does.

src/appmixer/ai/mistral/ExtractDocumentText/ExtractDocumentText.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+            const dataUri = `data:${contentType};base64,${base64}`;
+            document = IMAGE_TYPES.includes(contentType)
+                ? { type: 'image_url', image_url: dataUri }
+                : { type: 'document_url', document_url: dataUri };
+        } else {
+            const ext = path.extname(new URL(documentUrl).pathname).toLowerCase();
```
vtalas: ⚠️ **warning** · `behavior.unwrapped-error`

`new URL(documentUrl)` throws a raw `TypeError [ERR_INVALID_URL]` when the user supplies a malformed Document URL, so the component fails with an opaque platform error and is treated as a retryable fault rather than a user input mistake.

**Suggested fix:** Wrap the URL parse in a try/catch and rethrow as `new context.CancelError('Document URL must be a valid URL.')`, matching how parsePages reports bad input.

vtalas (commented):
## 🤖 Connector review

**0 error(s), 4 warning(s), 1 info** across 4 component(s).

This PR adds a new Mistral AI connector (appmixer.ai.mistral) with 10 components — SendPrompt, DescribeImages, GenerateEmbeddings, ExtractDocumentText, CreateTranscription, CreateSpeech, Moderate, ListModels, ListVoices and MakeApiCall — plus auth.js, lib.js, quota.js, bundle.json, module.json and four E2E test flows. I reviewed every component.json and behaviour file against the connector design rules, compared conventions against the sibling ai/* connectors (openai, claude, gemini, groq), and confirmed `appmixer connector validate --changed` passes all 22 validators with no failures or warnings. The connector is in good shape: it is well documented, the test-plan artifact shows the risky assumptions (CreateSpeech base64 audio, Moderate category keys, Mistral's plain-string image_url) were verified against live API responses, and all 10 components are covered by E2E flows. Findings are four low-to-moderate issues: a JPEG MIME fallback bug in ExtractDocumentText, missing error suppression on the two dynamic-source list components, a truncated cache key in lib.js, and an unwrapped URL parse; plus one test-coverage note.

|---|---|---|---|---|
| ⚠️ warning | appmixer.ai.mistral.ExtractDocumentText<br><sub>src/appmixer/ai/mistral/ExtractDocumentText/ExtractDocumentText.js:45</sub> | `behavior.invalid-mime-fallback` | The contentType fallback builds `image/${ext.substring(1)}`, which yields the invalid MIME type `image/jpg` for a `.jpg` file; that value is not in IMAGE_TYPES, so a JPEG whose stored contentType is missing is sent to the OCR API as `document_url` with a `data:image/jpg;...` URI instead of `image_url`. | Map extensions to real MIME types (`.jpg`/`.jpeg` -> `image/jpeg`) the same way DescribeImages.js:37 already does, rather than deriving the MIME type from the extension string. |
| ⚠️ warning | appmixer.ai.mistral.ListModels<br><sub>src/appmixer/ai/mistral/ListModels/ListModels.js:87</sub> | `source.error-suppression-missing` | ListModels and ListVoices back eight inspector dropdowns but neither passes an `isSource` sentinel in `source.data.properties` nor catches fetch errors, so an invalid key, an unsaved auth or a Mistral outage surfaces as an error popup every time a component's inspector is opened. | Pass `"properties": { "isSource": true }` in each dropdown's `source.data` and, when that flag is set, catch the request error and return an empty result instead of throwing — as ai/claude/ListModels/ListModels.js already does. |
| ⚠️ warning | appmixer.ai.mistral.ListVoices<br><sub>src/appmixer/ai/mistral/lib.js:46</sub> | `source.cache-key-collision` | The shared cache key uses only the last 8 characters of the API key, so two accounts whose keys happen to share that suffix would read each other's cached response — which for /audio/voices includes account-private custom voices. | Key the cache on a hash of the full credential, e.g. `crypto.createHash('sha256').update(JSON.stringify({ url, apiKey: context.auth.apiKey })).digest('hex')`, as the reference `callEndpointCached` does. |
| ⚠️ warning | appmixer.ai.mistral.ExtractDocumentText<br><sub>src/appmixer/ai/mistral/ExtractDocumentText/ExtractDocumentText.js:52</sub> | `behavior.unwrapped-error` | `new URL(documentUrl)` throws a raw `TypeError [ERR_INVALID_URL]` when the user supplies a malformed Document URL, so the component fails with an opaque platform error and is treated as a retryable fault rather than a user input mistake. | Wrap the URL parse in a try/catch and rethrow as `new context.CancelError('Document URL must be a valid URL.')`, matching how parsePages reports bad input. |
| ℹ️ info | appmixer.ai.mistral.CreateTranscription | `test.branch-not-covered` | The filepicker branches of CreateTranscription and ExtractDocumentText are never exercised — the documents-and-audio E2E flow uses only `fileUrl` and `documentUrl`, and the test plan notes the CLI harness does not mock loadFile/getFileInfo — leaving the multipart FormData path and the base64 data-URI/contentType-inference path untested. | Extend the documents-and-audio flow with an upload step so at least one run drives CreateTranscription and ExtractDocumentText through a real fileId, which would also have caught the `image/jpg` fallback above. |

_4 of these are also posted inline on the changed lines._

- `appmixer connector validate --changed` reports OK for all 22 validators (bundle-versions, component-schemas, makeapicall-standards, output-port-examples, quota-resources, required-inputs-asserted, ...) with no failures or warnings
- No prompt-injection or instruction-like content found in any source file, comment, artifact or test fixture; no hardcoded credentials and no attempt to read env vars or contact non-Mistral hosts
- Component names match their disk paths and the appmixer.ai.mistral.<Component> convention; labels, descriptions, icons, auth and quota blocks are present and consistent on all 10 components
- All required inputs are asserted in receive() with context.CancelError (Model/Prompt/Text/Voice, and the either-or checks in CreateTranscription and ExtractDocumentText)
- outPorts declare type + example on every leaf property; SendPrompt/Moderate dotted option paths line up with the shapes the behaviours actually emit (choices collapsed to choices[0]; the 11 live moderation category keys)
- ListModels and ListVoices correctly use lib.sendArrayOutput / lib.getOutputPortOptions with the `result` array field, and their generateOutputPortOptions source URLs carry ignoreAuth=true and pass in/outputType
- Dynamic source fetches are cached under context.lock + context.staticCache with context.config.listCacheTTL — the sanctioned unconditional-cache variant, appropriate here since ListModels backs seven dropdowns
- Model dropdowns are `type: "text"` with a source (never `select`), each has a sensible defaultValue, and they are capability-filtered so a component only offers models its endpoint accepts
- No async/long-running job shape: every component returns its actual result synchronously, none exposes a callback-URL input, and there is no companion poll-for-status component
- quota.js uses the standard `window-sliding` rule shape and its `requests` resource matches the `resources` reference in all 10 component.json quota blocks
- All 10 components appear in at least one of the four E2E test flows
- SendPrompt's conversation-state handling and MakeApiCall's URL/key-value handling follow the established ai/groq and ai/openai patterns verbatim, keeping the connector family consistent

</details>

petrolivka:
Superseded by #1253 — same commits, re-opened from a fork I can push to (the head branch here lives on `apx-vero`, which I don't have write access to). Please continue review there.

petrolivka:
Superseded by #1253 — same commits, re-opened from a fork I can push to (this PR's head branch lives on `apx-vero`, which I don't have write access to). Please continue review there. I can't close this one myself; a maintainer would need to.

PR #1219, opened 2026-08-24
validators: warn when a writable field cannot be read back

apx-vero:
Adds a validator for a gap nothing in the repo currently covers: **a field you can write but never read back.**

## The gap

All 26 existing connector validators are static shape checks — I counted, **none of them makes a live API call**. E2E flows do touch the real API, but they only assert the two or three fields the flow author happened to pick. So a field can be spelled correctly, carry a valid schema with an `example`, pass every gate, and still be invisible to the variable picker.

That is exactly what happened in the Cliniko connector: `Create Patient` accepted `medicare`, Cliniko stored it, and the Patient output never declared it. A flow could set the number and never read it back.

## What it does

For every `Create<Entity>` / `Update<Entity>` it finds the matching reader — `Get<Entity>`, or the writer's own `out` port when that returns the created entity — and reports inputs with no counterpart in the output.

112 fields across the repo. Spot-checked survivors are real:

| Component | Write-only | Reality |
|---|---|---|
| `github/list/UpdateIssue` | `assignees`, `labels` | declares singular `assignee.*` and no labels; the API returns both |
| `stripe/core/CreatePaymentIntent` | `description`, `customer`, `captureMethod` | all returned by Stripe |
| `vapi/core/CreateAssistant` | `firstMessage`, `model`, `voice` | all returned |
| `freshsales/core/CreateTask` | `priority`, `contact_id`, `deal_id` | all returned |

## Warning, not a failure — and no threshold

Reported via `addWarning`, which never fails CI, and deliberately **not** given a `.thresholds.json` entry.

The input and output sides of an API legitimately differ in shape, and the heuristic cannot always tell that apart from a real gap. A hard gate would be actively harmful here: pre-commit validates changed files **strictly, ignoring thresholds**, so one false positive would block a commit that merely touched a neighbouring line. As a warning it surfaces in `--connector <name>` runs — the workflow connector authors already use — and costs nobody anything.

## Noise reduction, each step measured

| Filter | Warnings |
|---|---|
| naive name match | 188 |
| descend into nested object / array-item schemas (`patient_phone_numbers[].phone_type` counts as readable) | 172 |
| require the entity to be retrievable — a `Get`/`Find`/`List` sibling — dropping operations merely *named* `Create*` (`openai CreateSpeech`, `groq CreateTranscription`) whose inputs are parameters, not entity fields | 117 |
| accept a differently-named counterpart via containment (`deal_name` → `name`, `dueString` → `due`) | — |
| skip the record's identifier, normalised, so `productId`, `product_id` and github's bare `issue` are all recognised | 112 |

Known remaining false-positive classes are named in the file header rather than hidden: request-scoping parameters (`xero tenantId`, `googleTasks tasklist`) and action modifiers that are not stored (`lemonsqueezy invoiceImmediately`). The header says plainly that a report is a prompt to look, not a verdict.

## Verification

`node scripts/validate.js --connector cliniko` → 0 failures, 0 warnings (the `medicare` gap it originally found is fixed in `dev`). Repo-wide run unchanged for every other validator. eslint clean.

https://claude.ai/code/session_011g7A4beDMYJu9kqmNu26Pp

### DIFF
--- a/scripts/validators/writable-fields-readable.js
+++ b/scripts/validators/writable-fields-readable.js
@@ -1,0 +1,257 @@
+'use strict';
+
+// What this validator checks
+// --------------------------
+// A field a user can WRITE should be a field they can READ BACK. For every
+// Create<Entity> / Update<Entity> component this validator finds the matching
+// reader — Get<Entity>, or the writer's own `out` port when it returns the
+// created entity — and reports inputs that have no counterpart in the reader's
+// output schema.
+//
+// Why
+// ---
+// Such a field is write-only from the flow author's point of view: the inspector
+// offers it, the API stores it, and the variable picker never shows it again, so
+// nothing downstream can branch on what was just written. Nothing else catches
+// this. The field is spelled correctly and carries a valid schema, so the static
+// validators are happy; and E2E flows only assert the two or three fields the
+// flow author happened to pick. Real case: cliniko Create Patient accepted
+// `medicare` and the Patient output never returned it.
+//
+// WARNING, not a failure
+// ----------------------
+// Reported through addWarning because the input and output sides of an API
+// legitimately differ in shape, and the heuristic below cannot always tell that
+// apart from a genuine gap. Measured over this repo, roughly half of the reports
+// are structural rather than bugs:
+//
+//   - several inputs composing one output structure
+//     (todoist dueString/dueDate/dueDatetime -> due, cliniko phoneNumber ->
+//     patient_phone_numbers[])
+//   - request-scoping parameters that are not entity fields at all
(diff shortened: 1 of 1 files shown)


apx-vero:
Closing in favour of a single source of truth: the rule now lives in the CLI's connector-validate suite (Appmixer-ai/appmixer-cli#422, commit ceeaf6e) — same file, same behaviour, verified firing over this workspace through `appmixer connector validate`.

Direction agreed with Vláďa: this repo's CI/pre-commit will consume the validator suite from appmixer-cli's dev branch (`appmixer connector validate --changed`) instead of duplicating rules in `scripts/validators/`. Wiring PR follows.

PR #1226, opened 2026-08-25
exa: new connector (6 components, API key auth)

petrolivka:
## Summary

New connector `appmixer.ai.exa` — [Exa](https://exa.ai), a web search API built for AI agents. Semantic and keyword web search, clean LLM-ready content extraction, semantic "find similar pages", a cited AI answer endpoint, a polling trigger, and a generic authorized API call.

Initial release, so `bundle.json` stays at **1.0.0**.

## Components

| Component | Label | Endpoint | Shape |
|---|---|---|---|
| `FindSearchResults` | Search Web | `POST /search` | Find, `outputType` |
| `FindSimilarLinks` | Find Similar Links | `POST /findSimilar` | Find, `outputType` |
| `FindExtractedPages` | Extract Page Contents | `POST /contents` | Find, `outputType` |
| `AnswerQuestion` | Answer Question | `POST /answer` | Action, static schema |
| `MakeApiCall` | Make API Call | any Exa endpoint | Action, static schema |
| `NewSearchResult` | New Search Result | `POST /search` | Polling trigger with `test()` |

## Auth and quota

**API key**, sent as `x-api-key` (Exa also accepts `Authorization: Bearer`).

`validate()` uses `POST /contents` with an empty body: a valid key is rejected with `400 INVALID_REQUEST_BODY` (the key authenticated, the body did not), an invalid key with `401 INVALID_API_KEY`. Every Exa endpoint that returns data bills the account, so this is the only way to check a key for free.

`quota.js` throttles to 10 QPS — Exa's lowest documented ceiling (`/search` and `/answer` are 10 QPS; `/contents` allows 100).

`MakeApiCall` pins the target host to `https://api.exa.ai` through `lib.resolveApiUrl`, so an absolute URL aimed at a third-party host cannot leak the account's API key.

## Test evidence

**Workspace validator** — `node scripts/validate.js --connector ai/exa`: **0 failures, 0 warnings** across 26 validators.

**Lint** — `eslint src/appmixer/ai/exa/`: clean.

**CLI component tests** — `appmixer test component`, **6/6 passed** against the live Exa API:

| Component | Result |
|---|---|
| `FindSearchResults` | 3 results on `out` |
| `FindExtractedPages` | full page text extracted on `out` |
| `FindSimilarLinks` | 3 scored similar pages on `out` |
| `AnswerQuestion` | answer + 6 citations on `out` |
| `MakeApiCall` | `statusCode: 200` on `out` |
| `NewSearchResult` | `--test` emitted one result; a live `tick()` loop confirmed the first run saves a baseline and emits nothing, and later ticks emit only unseen pages |

**E2E on `api-dev-automated-00001.dev.appmixer.ai`** — `appmixer e2e validate` clean (26 validators, 2 flows), both flows imported and ran green:

| Flow | Flow ID | Result |
|---|---|---|
| [E2E Exa - Search, Extract and Find Similar](src/appmixer/ai/exa/artifacts/test-flows/test-flow-search-extract-similar.json) | `f6b20f14-6585-4e94-8592-b759de209ff7` | PASSED — 3 asserts |
| [E2E Exa - Answer and API Call](src/appmixer/ai/exa/artifacts/test-flows/test-flow-answer-apicall.json) | `b2c578b6-6454-4c41-9626-ce6700178927` | PASSED — 2 asserts |

Verified in the E2E result store (`appmixer e2e results -c appmixer:ai:exa`), not just the runner's PASSED line: **0 failed records**, and each of the 5 Assert components recorded a success clause with no failure clauses.

## Notes for reviewers

- **`NewSearchResult` is not covered by an E2E flow** — it is a flow-starting trigger. It is covered by the CLI `--test` run and the live `tick()` loop described above. Worth a look if you want E2E coverage for it.
- The first `appmixer e2e import` failed with `lib.getOutputPortOptions is not a function` on `FindExtractedPages` only, while its two sibling `outputType` components resolved fine and the same call worked locally. A second `appmixer publish` cleared it — the partial-publish behaviour this instance has shown before, not a connector defect. Flagging it in case it recurs.
- `lib.js` follows the canonical `sendArrayOutput` / `getOutputPortOptions` helpers; the array output field is `result`.
- No `package.json` — the connector needs no npm dependencies, everything goes through `context.httpRequest`.

https://claude.ai/code/session_017qK8NDL5GNuqSVDGzSJALu


petrolivka:
- https://dev-automated-00001.dev.appmixer.ai/designer/f6b20f14-6585-4e94-8592-b759de209ff7
- https://dev-automated-00001.dev.appmixer.ai/designer/b2c578b6-6454-4c41-9626-ce6700178927

vtalas (commented):
## 🤖 Connector review

**0 error(s), 4 warning(s), 2 info** across 4 component(s).

This PR adds a new Exa connector (appmixer/ai/exa) with API-key auth and six components: three outputType-based Find components (FindSearchResults, FindExtractedPages, FindSimilarLinks), an AnswerQuestion action, a generic MakeApiCall, and a NewSearchResult polling trigger, plus connector-level auth.js, lib.js, quota.js, bundle.json, module.json and two E2E flows. I reviewed every component.json and behavior file against the connector standards, compared lib.js to the canonical template, checked the trigger's tick()/test() contract, audited MakeApiCall's SSRF/key-leak surface, and cross-read /tmp/connector-validate.txt. The connector is in good shape overall — auth, quota, outputType handling, notFound routing and output-port schemas are all correct, and MakeApiCall correctly pins the request origin so the API key cannot be sent to a foreign host. The findings are the three validator warnings about variable-bound required inputs in the dynamic outPort sources (which degrade the designer's variable picker to Raw Output) and a defensive empty-array guard that was dropped from the canonical lib.js toCsv helper; no errors were found.

|---|---|---|---|---|
| ⚠️ warning | appmixer.ai.exa.FindSearchResults<br><sub>src/appmixer/ai/exa/FindSearchResults/component.json:241</sub> | `validate.dynamic-outport-required-inputs` | The outPort source maps the required input `in/query` from the live flow value `inputs/in/query`, which makes the designer's required-input validation reject the schema call so the variable picker degrades to Raw Output. | Send the literal "dummy" for `in/query` in `source.data.messages` — the value does not shape the generated schema. |
| ⚠️ warning | appmixer.ai.exa.FindExtractedPages<br><sub>src/appmixer/ai/exa/FindExtractedPages/component.json:141</sub> | `validate.dynamic-outport-required-inputs` | The outPort source maps the required input `in/urls` from the live flow value `inputs/in/urls`, which makes the designer's required-input validation reject the schema call so the variable picker degrades to Raw Output. | Send the literal "dummy" for `in/urls` in `source.data.messages` — the value does not shape the generated schema. |
| ⚠️ warning | appmixer.ai.exa.FindSimilarLinks<br><sub>src/appmixer/ai/exa/FindSimilarLinks/component.json:209</sub> | `validate.dynamic-outport-required-inputs` | The outPort source maps the required input `in/url` from the live flow value `inputs/in/url`, which makes the designer's required-input validation reject the schema call so the variable picker degrades to Raw Output. | Send the literal "dummy" for `in/url` in `source.data.messages` — the value does not shape the generated schema. |
| ⚠️ warning | appmixer.ai.exa<br><sub>src/appmixer/ai/exa/lib.js:221</sub> | `lib.tocsv-empty-guard-missing` | lib.js `toCsv` drops the canonical template's empty-array guard and indexes `array[0]` directly, so any future caller of `sendArrayOutput({ outputType: 'file' })` with zero records throws a TypeError that kills the flow instead of writing an empty file (today's three callers happen to route empty results to notFound first, so it is latent). | Restore the canonical guard: `if (!array.length) { return ''; }` before `Object.keys(array[0])`. |
| ℹ️ info | appmixer.ai.exa.FindSearchResults<br><sub>src/appmixer/ai/exa/FindSearchResults/component.json:100</sub> | `inspector.enum-unverified` | The Search Type options offer `instant`, `deep-lite` and `deep` while omitting Exa's long-standing `neural` and `keyword` types, and the sibling NewSearchResult trigger exposes a different subset for the same `/search` `type` parameter — if any of these values is not in Exa's documented enum the call fails with a 400 at runtime. | Cross-check the option values against Exa's documented `/search` `type` enum and align the two components' lists (or document why the trigger deliberately restricts to the cheap types). |
| ℹ️ info | appmixer.ai.exa<br><sub>src/appmixer/ai/exa/artifacts/ai-artifacts/pipeline-state.json:34</sub> | `artifact.stale-validator-result` | pipeline-state.json records the validator result as "0 failures, 0 warnings", but `connector validate --changed` reports three dynamic-outport-required-inputs warnings on this connector, so the committed artifact understates the actual validation state. | Update the recorded validator result once the three warnings above are fixed, so the artifact reflects the real run. |

_6 of these are also posted inline on the changed lines._

- auth.js uses the apiKey type with the x-api-key header, matching lib.makeRequest and MakeApiCall; validate() distinguishes 401 (bad key) from 400 (valid key, empty body) and rethrows transient errors instead of misreporting them as a bad key
- MakeApiCall pins the target to https://api.exa.ai via lib.resolveApiUrl, rejecting protocol-relative (//host), credentialed (user:pass@) and foreign-origin URLs, so the account API key cannot be leaked to a third-party host
- MakeApiCall spreads user headers before x-api-key, so a user-supplied header cannot override the auth header
- All three Find components use lib.sendArrayOutput / lib.getOutputPortOptions, emit the array under the required `result` field with `count`, and route empty result sets to the notFound port
- generateOutputPortOptions short-circuits before any HTTP call in all three Find components, so the self-referencing source.url never hits the Exa API (dynamic-source caching requirement does not apply)
- Every outPort leaf property carries type + title + a realistic example; schema and options are never mixed on the same port
- NewSearchResult is a correct polling trigger: properties (not inPorts), trigger+tick flags, first-run baseline so existing pages do not fire, bounded known-id window, and a test() that reuses the shared search() helper, honors context.properties filters, writes no state, emits exactly one item and throws when no example exists
- Required inputs are validated with context.CancelError in every component (query, urls, url, method)
- quota.js and the per-component quota blocks follow the repo convention and resolve to the declared `requests` resource; scope matches sibling connectors
- Published-date inputs use the date-time inspector type with enableTime: false, not text
- Not an async/job-submitting API — no callback URL is exposed as an input and no completion-path (self-callback or continuation chain) requirement applies
- E2E flows follow OnStart -> SetVariable -> components -> Assert -> AfterAll -> ProcessE2EResults and use only equal/regex/notEmpty assertions; 5 of 6 components are covered, with the NewSearchResult trigger verified via CLI test() instead (documented in pipeline-state.json)
- No prompt-injection, credential-exfiltration or unsafe-execution patterns in any source, artifact, test flow or icon file

</details>

apx-vero:
Superseded by #1256 — please close this one when you get a chance, @petrolivka.

I picked this connector up to get it release-ready, which needed a tracker issue and a fresh E2E run. Adding commits here wasn't possible from this account (the branch is on your fork), so your five commits are rebased onto current `dev` in #1256 with authorship preserved, plus one commit on top.

What that commit changes, after checking the option lists against the live API rather than the docs:

- **`category` was wrong and incomplete.** The API doesn't validate this field — an unknown value is accepted and silently ignored on `/search`, so a bad option degrades to unfiltered results instead of erroring. `publication` does work (Exa resolves it to `research paper`), but `research paper`, `pdf` and `github` were missing and return visibly different result sets.
- **`notFound` ports had no schema**, so a flow couldn't bind the echoed `query`/`url`/`urls` — `appmixer e2e import` rejected an assert on `.notFound.query` for exactly that reason.
- **`FindExtractedPages` never declared `subpages`** although it exposes the Subpages inputs.
- A third E2E flow covers the `array` outputType and the `notFound` port; the two existing flows only drove `outputType: "first"` on `out`.
- Three stray `value` keys on regex assert clauses are dropped — `appmixer e2e validate` fails on them.

Your `searchType` list was checked too and is a valid subset, and the SSRF pinning in `resolveApiUrl` held up against 15 hostile inputs. Nice work on that part.

E2E is green — 3 flows, 7 asserts, 0 failures on `api-dev-automated-00001`, with a negative control proving the suite can fail. Full report on Appmixer-ai/appmixer-components#2833.

One thing that will bite the next person: `appmixer e2e import` can't resolve the account for nested `ai/*` refs and silently skips the binding, after which `e2e run` fails preflight with "No valid account bound". Pass `-a <accountId>`.

PR #1227, opened 2026-08-25
Hugging Face connector (8 components, API key auth)

petrolivka:
## Summary

New **Hugging Face** connector (`appmixer:ai:huggingface`), placed under the `ai/` sub-vendor alongside `openai`, `mistral` and `exa`. API-key auth with a User Access Token, validated against the free Hub `whoami-v2` endpoint.

All eight components are verified against the live Hugging Face API, and both E2E flows pass on the dev instance.

## Components

| Component | Type | Endpoint |
|---|---|---|
| `SendPrompt` | Action | `POST router.huggingface.co/v1/chat/completions` (OpenAI-compatible) |
| `CreateEmbeddings` | Action | `feature-extraction` task route; mean-pools token vectors, returns one vector + `dimensions` |
| `ClassifyText` | Action | task route; unwraps `[[{label,score}]]`, returns top label + all labels sorted |
| `SummarizeText` | Action | task route; unwraps `[{summary_text}]` |
| `GenerateImage` | Action | Inference Provider route; saves the image to file storage, returns `fileId` |
| `FindModels` | Find (`outputType`, `notFound`) | Hub `GET /api/models` |
| `GetModel` | Get | Hub `GET /api/models/{repo_id}` |
| `MakeApiCall` | Action | Escape hatch, origin-pinned to the Hub and router hosts |

## Auth and quota

User Access Token sent as `Authorization: Bearer hf_...`. `requestProfileInfo` labels the account with the real user/org name and falls back to a masked token on error, so a valid token is never blocked from connecting. `MakeApiCall` resolves its target against an allow-list of the Hub and router origins so the token cannot leak to a third-party host.

## Test evidence

**Validator** — `node scripts/validate.js --connector ai/huggingface`: 0 failures, 0 warnings (25 rules).
**Lint** — eslint 8.57.1: clean.

**CLI component tests — 8/8 against the live API:**

| Component | Result |
|---|---|
| `FindModels` | PASS — `outputType: first`; `filter` + `author` repeated-param serialization verified |
| `GetModel` | PASS |
| `MakeApiCall` | PASS — HTTP 200 |
| `CreateEmbeddings` | PASS — 384 dimensions |
| `ClassifyText` | PASS — `POSITIVE` @ 0.99988 |
| `SummarizeText` | PASS |
| `SendPrompt` | PASS — full chat-completion shape with token counts |
| `GenerateImage` | PASS after the fix below — 221 KB image saved to file storage |

**E2E — 2/2 flows on `api-dev-automated-00001.dev.appmixer.ai`, 13 assertions, 0 failures**, read from the results store rather than the runner's PASSED line:

- `E2E Hugging Face - Hub Models and API Call` (`ff5fbf6d`) — 5 assertions
- `E2E Hugging Face - Inference Tasks` (`af10765b`) — 8 assertions

Flow JSONs live in [`src/appmixer/ai/huggingface/artifacts/test-flows/`](src/appmixer/ai/huggingface/artifacts/test-flows/).

## Notes for reviewers

**`GenerateImage` needed a real fix and is the part most worth reviewing.** As first written it could never have worked: it routed text-to-image through the legacy `hf-inference` provider, which no longer serves that task — popular checkpoints answer `HTTP 410 "The requested model is deprecated and no longer supported by provider hf-inference"`, others `400 "Model not supported"`. The router exposes no OpenAI-compatible `/v1/images/generations` endpoint either (404).

It now routes through a named Inference Provider: a new `provider` input (default `fal-ai`) builds `{ROUTER_BASE_URL}/{provider}/{model}`, with the model path deliberately **not** repo-encoded so its slashes survive into the URL. The `model` input is therefore the provider's own model path (`fal-ai/flux/schnell`), not always the Hub repo ID — this is called out in the tooltip.

Two smaller points in the same component:

- Because the request uses `responseType: arraybuffer`, error bodies previously surfaced as an unreadable array of raw byte values. They are now decoded to text, so a dead model reports `Provider hf-inference cannot serve model X: Model not supported by provider hf-inference`.
- Response handling covers `images[].url` (fal.ai), `images[]` as a base64 string, `output[]`, `data[].url`, `data[].b64_json` and `image_url`.

Other things worth knowing:

- The router silently remaps `meta-llama/Llama-3.1-8B-Instruct` to `Meta-Llama-3.1-8B-Instruct-Turbo` in the response, so E2E never asserts equality on the returned model id.
- `SendPrompt` at `temperature: 0` is still not word-exact, so the E2E assertions use factual substrings (`Lisbon`, `Eiffel`) rather than exact echoes.
- `facebook/bart-large-cnn` can need a cold start of roughly two minutes on the first call.
- Initial release, so `bundle.json` stays at 1.0.0 with no changelog entry.


vtalas (commented):

**First, the good news:** I reproduced the validator, lint and e2e assertion counts independently and **they match your PR description exactly** — validator 0 failures / 0 warnings across all 25 rules, eslint clean, 13 e2e assertions. Nothing here disputes your test evidence; these are findings the 25 static rules don't express.

**1 error, 2 warnings, 5 info.** Inline comments cover the errors and warnings. Info items are listed below.

### error · `repo.stray-artifact`

*(In the body rather than inline — GitHub can't anchor a comment to a binary file.)*

`src/appmixer/ai/huggingface/GenerateImage/e2e-test.png` is a git-tracked **221 040-byte binary** with nothing referencing it — `grep -rn "e2e-test" src/appmixer/ai/huggingface` returns zero hits. It's the live E2E output, per `artifacts/ai-artifacts/pipeline-state.json:80` (*"PASS after fix — live; 221 KB image saved to file storage via fal-ai provider"*), added by commit `bc8c9875` ("CLI tests 8/8 — fix GenerateImage provider routing") as a side effect of the test run.

It is the **only non-icon PNG inside any component directory in the whole repo** — the six other committed non-icon images are all documentation or UI assets (`apigee/architecture.png`, the Google sign-in button, the `utils/*` loaders). So `appmixer pack` will ship 221 KB of test residue inside the bundle.

```
git rm src/appmixer/ai/huggingface/GenerateImage/e2e-test.png
```

If a sample output is worth keeping, it belongs under `artifacts/`, not the component folder. Worth noting you already cleaned up the equivalents in the sibling PRs ("brightdata: remove accidentally committed build archive", "airtop: drop the stray icon.svg") — this one just survived.

### Info (5)

- **Silent embedding degradation** — `CreateEmbeddings.js:79-81`: when the per-token rows are ragged, the code returns **the first token's** vector instead of a pooled sentence embedding. The out port then reports `dimensions: embedding.length` with no signal that pooling was skipped, so a downstream similarity search gets a semantically wrong vector that looks perfectly valid. Suggest a `context.log` at minimum, or a `CancelError` saying the model isn't a sentence-level feature-extraction model.
- **Hardcoded `image/png`** — `GenerateImage.js:96-98` assumes PNG for every inline-base64 response regardless of actual format, so a base64 JPEG gets a `.png` filename (`:106-109`) and a wrong `contentType` (`:116`). The remote-URL branch reads the real header at `:93-95`. Sniffing magic bytes would fix it.
- **Input order in MakeApiCall** — `api` ("API Host") is `index: 6` so it renders last, after Headers, yet it decides which origin a relative path resolves against (`MakeApiCall.js:33`). `url` is `index: 1`. The user picks the path before being shown what it's relative to.
- **Undeclared `notFound` field** — `FindModels.js:111` sends `{ search: search || null }`, but the port is declared with no `schema` and no `options`, so the field never appears in the variable picker. The canonical example sends `{}` (`references/examples/find-tasks/FindTasks.js:41`).
- **Inconsistent numeric coercion** — `SendPrompt.js:49-51` passes `max_tokens` raw while `:52-57` wraps `temperature` and `topP` in `Number()`; same split in `ClassifyText.js:20-22` for `top_k`. `GenerateImage.js:141-152` coerces everything. An expression-bound value arriving as a string would be sent as `"max_tokens": "500"`.

---

vtalas (approved):

PR #1228, opened 2026-08-25
Airtop connector (10 components, API key auth)

petrolivka:
## Summary

New **Airtop** connector (`appmixer:airtop`) — cloud browser automation. API-key auth, 10 components covering the full session → window → interact → terminate lifecycle.

All ten components are verified against the live Airtop API, and both E2E flows pass on the dev instance.

## Components

| Component | Type | What it does |
|---|---|---|
| `CreateSession` | Create | `POST /sessions` — profile, idle timeout, captcha solving, proxy, recording, extensions |
| `GetSession` | Get | `GET /sessions/{id}` — status, usage minutes, CDP URLs |
| `FindSessions` | Find (`outputType`, `notFound`) | `GET /sessions`, filterable by status |
| `TerminateSession` | Delete | `DELETE /sessions/{id}` → returns `{}` |
| `CreateWindow` | Create | `POST /sessions/{id}/windows` — echoes `sessionId` so a flow can chain |
| `QueryPage` | Action | `page-query` — natural-language question, optional `outputSchema` for structured JSON |
| `ClickElement` | Action | `click` — natural-language element description |
| `TypeText` | Action | `type` — text, target field, Enter/clear/wait |
| `TakeScreenshot` | Action | base64 data URL or signed download URL |
| `MakeApiCall` | Action | Authorized escape hatch |

Built from Airtop's published OpenAPI spec (`https://docs.airtop.ai/openapi.json`), not from guesswork — every endpoint, field name and enum comes from it.

## Auth and error handling

API key sent as `Authorization: Bearer <key>`, validated with a read-only `GET /sessions?limit=1` that spends no session minutes.

`lib.js` carries the canonical `sendArrayOutput` / `getOutputPortOptions` helpers (array field `result`), plus `unwrap()` for Airtop's `{data, meta, errors, warnings}` envelope and `aiResult()`. The latter matters: **Airtop answers 2xx even when an AI operation fails**, so `meta.status` is the real success signal — `aiResult` raises a `CancelError` on `failure` and on a populated `errors[]`.

## Test evidence

**Validator** — 0 failures, 1 warning. **Lint** — clean. **`e2e validate`** — passed (26 validators).

**CLI component tests — 10/10 against the live API**, including both `FindSessions` ports (`notFound` with no sessions, `out` with a populated `result[]`) and `TerminateSession` returning `{}`.

**E2E — 2/2 flows on `api-dev-automated-00001.dev.appmixer.ai`, 10 assertions, 0 failures**, read from the results store rather than the runner's PASSED line:

- `E2E Airtop - Session Lifecycle` (`6a6a200b`) — 4 assertions
- `E2E Airtop - Browser Interaction` (`7852000a`) — 6 assertions

Confirmed no sessions were left running after the suite, so nothing keeps billing.

## Notes for reviewers

**`CreateSession` needed a real fix and is the main thing to review.** It emitted the session as soon as `POST /sessions` returned, while Airtop still reports `status: "initializing"`. A session is unusable until `running` — `CreateWindow` and every page operation answer **404** before then. This passed its CLI test only because `GetSession` happened to run in between; in any realistic flow (create session → open window) it failed every time, which is exactly how the E2E flow caught it.

`CreateSession` now polls `GET /sessions/{id}` every 2 s until the session reports `running`, with a 120 s ceiling and a `CancelError` if the session ends first.

The same race broke the session-lifecycle E2E flow from the other side: it filtered `FindSessions` on `status: "running"` while the session was still initializing, so the component took its `notFound` port. Its `Assert` never fired, `AfterAll` never completed, and **`TerminateSession` never ran — leaking a billed session on every run.** The filter was removed.

Two deliberate non-fixes:

- **`writable-fields-readable` warning on `CreateSession`** — its inputs have no counterpart in `GetSession`'s output. Airtop only echoes them inside a `configuration` object its own spec marks **deprecated**, so surfacing it would bake a deprecated field into the contract.
- **No dropdown `source` on `sessionId`/`windowId`** — sessions are ephemeral (10 min idle default) and normally come from `CreateSession` in the same flow, so a dropdown would be near-always stale.

Operational findings worth knowing:

- **Measured credit costs:** `QueryPage` 15, `TypeText` 27, **`ClickElement` 485**. Clicking is by far the most expensive operation — worth keeping out of frequently-run flows.
- The first flow start after *every* publish fails with `TokenError: Missing 'type' property` and succeeds on re-run. This is an instance artifact, not a connector defect — observed identically after two separate publishes, with a clean pass whenever no publish intervened.
- The test account is on a free plan (`recordingState: "free_plan_disabled"`), so `CreateSession`'s `record` option could not be exercised.
- The session-lifecycle assert on `FindSessions` is weak — it asserts the whole `result` array is notEmpty, which stringifies to `[object Object],...`. It passes, but a future revision should assert a specific field.

Follow-up candidates for 1.1.0: Close Window, Load URL, Hover, Scroll, Monitor, PaginatedExtraction, and save-profile-on-termination. Airtop publishes no rate limits, so `quota.js` uses a conservative 10 req/s per user. Initial release, so `bundle.json` stays at 1.0.0.


vtalas (commented):

**First, the good news:** I reproduced the validator, lint and e2e assertion counts independently and **they match your PR description exactly** — validator 0 failures / 1 warning (the `writable-fields-readable` one you documented), eslint clean, 10 e2e assertions. Nothing here disputes your test evidence; these are findings the 25 static rules don't express.

**1 error, 5 warnings, 11 info.** Inline comments cover the errors and warnings. Info items are listed below.

### Info (11)

- **Blocking sleep** — `CreateSession.js:96` holds a worker for up to 120 s instead of a `context.setTimeout` continuation. Deliberately *not* raised higher: `ai/agentcore/CreateMemory` does the same thing on purpose, and the continuation floor of 60 s is too coarse for a ~5–20 s readiness wait. Worth considering a `waitForRunning` toggle like agentcore's.
- **Copy-pasted examples** — `ClickElement/component.json:122` and `TypeText/component.json:125` both carry QueryPage's example verbatim (`"The top result is …"`). A click or a keystroke doesn't return a search summary; this misleads anyone composing downstream expressions.
- **`pipeline-state.json` contradicts itself** — line 27 says `"status": "BLOCKED"` for cli-tests while lines 37 and 147 say PASS, and line 20 claims `icon.svg in place` although commit `32ba23c3` deleted it. Shipping `artifacts/` is normal (90 of 160 connectors do), but a reviewer reading this one is actively misled.
- **`body.warnings` never read** — the `unwrap` docblock (`lib.js:55-56`) says Airtop reports non-fatal problems in `errors`/`warnings`, but only `body.errors` is ever read. Either surface them via `context.log` or drop the claim from the docblock.
- **`credits` missing on TakeScreenshot** — `aiResult` computes it (`lib.js:97`) and the other three AI components declare and emit it, but `TakeScreenshot/component.json:113-129` omits it. Screenshots cost credits too.
- **`CreateWindow` skips the `meta.status` check** — it uses `lib.unwrap`, which checks `errors[]` but not `meta.status`, unlike the four components routed through `aiResult`. Its own tooltip admits "On timeout the window is still created", so a navigation that never completed reports clean success.
- **`CreateWindow` unverified output fields** — `targetId`, `title` and `url` are declared with realistic examples but nothing verifies they exist; the e2e flow asserts only `windowId`. If they aren't in the create-window response, three declared variables resolve to `undefined` at runtime.
- **Duplicated default URL** — `https://www.google.com` is hardcoded in both `CreateWindow/component.json:335` and `CreateWindow.js:22`. A user who deliberately clears the field still lands on Google.
- **No `source` on `sessionId` inputs** — seven components take a bare `text` sessionId. Defensible as designed (sessions are ephemeral and normally chain from CreateSession), so this is a suggestion only. If added, `FindSessions` would then need `staticCache` + `lock`.
- **No error context wrapping** — `lib.apiRequest` never inspects `error.response`, so a 401/404/429 surfaces as a raw axios error with no hint which session or window was involved. Siblings are split on this; `cliniko/lib.js:146-190` is the hardened precedent.
- **Screenshot not persisted as a file** — with the default `format: base64` the whole image rides in the flow message. `08-best-practices.md` prefers `context.saveFileStream()` + `fileId`. No sibling forces this for screenshots, hence info only.

---

vtalas (approved):

PR #1229, opened 2026-08-25
Bright Data connector (7 components, API token auth)

petrolivka:
## Summary

New **Bright Data** connector (`appmixer:brightdata`) — web data platform: unblock and scrape any public page via the Unlocker API, pull parsed search-engine results via the SERP API, and drive the Web Scraper Datasets API (trigger a collection, poll the snapshot, download records).

All seven components are verified against the live Bright Data API, and both E2E flows pass on the dev instance.

## Components

| Component | Type | Endpoint |
|---|---|---|
| `ScrapeUrl` | Action | `POST /request` — page as HTML or Markdown |
| `FindSearchResults` | Find (`outputType`, `notFound`) | `POST /request` with `brd_json=1` — parsed `organic` results |
| `TriggerDatasetCollection` | Create | `POST /datasets/v3/trigger` → `snapshotId` |
| `GetSnapshotStatus` | Get | `GET /datasets/v3/progress/{id}` + derived `isReady` |
| `FindSnapshotRecords` | Find (`outputType`, `notFound`) | `GET /datasets/v3/snapshot/{id}` |
| `ListZones` | List (`outputType`) | `GET /zone/get_active_zones` |
| `MakeApiCall` | Action | Escape hatch, origin-pinned to `api.brightdata.com` |

**Auth:** API token as `Authorization: Bearer <token>`, validated against the free `GET /zone/get_active_zones`. `requestProfileInfo` is a pure function returning a masked key — no network call that could block a valid token from connecting.

## Test evidence

**Validator** — 0 failures, 0 warnings. **Lint** — clean. **`e2e validate`** — passed (26 validators).

**CLI component tests — 7/7 against the live API**, including `ScrapeUrl`'s `dataFormat: markdown` conversion and `FindSearchResults` returning real ranked Google organic results.

**E2E — 2/2 flows on `api-dev-automated-00001.dev.appmixer.ai`, 12 assertions, 0 failures**, read from the results store rather than the runner's PASSED line:

- `E2E Bright Data - Zones, Scrape, Search and API Call` (`758ddc7a`) — 7 assertions
- `E2E Bright Data - Dataset Snapshot Lifecycle` (`0622bb75`) — 5 assertions

## Notes for reviewers

**Two E2E flow bugs were found and fixed by running them, not by reading them:**

- The dataset flow asserted the snapshot id against regex **`^s_`**, which can never match. Real ids are `sd_xxxx`, so `^s_` demands an underscore exactly where a `d` is. It would have failed on every run. Corrected to `^sd_`.
- The same flow waited **5 minutes** for the snapshot. Measured readiness is a few seconds, so the wait was cut to `1m` — the old value burned wall-clock and crept toward the runner's 480 s timeout for no benefit.

**A documented API contradiction turned out not to matter.** Two Bright Data doc pages disagree on the `TriggerDatasetCollection` body — a plain array versus `{"input": [...]}`. Probed directly: **both return 200 with a `snapshot_id`**. A deliberately malformed control body returns `400 validation_error` with per-field detail, which confirms the API is not silently ignoring the payload in either case. The connector sends the plain array; the other shape would have worked too.

**Terminology, worth knowing when reviewing the `zone` input.** Bright Data has deprecated "zone" as a *dashboard* term — the control panel now calls these Proxies/APIs under *Proxies & Scraping Infrastructure*. The **wire parameter is still `zone`**: the API answers `zone "x" not found`, and `GET /zone/get_active_zones` is live. The field therefore stays, but its label is now **"Zone (Proxy / API name)"** and the tooltips point at where a user can actually find the value today. `ScrapeUrl` needs an **Unlocker API**; `FindSearchResults` needs a **SERP API**; the name you give each is the value.

Two operational findings:

- `GET /status` reporting `can_make_requests: false` / `auth_fail_reason: zone_not_found` governs the **Unlocker and SERP proxy products only**. The Datasets API is entirely unaffected — the dataset chain passes on an account that flag calls inactive. Worth not misreading as a dead account.
- The test token is scoped: `GET /customer/balance` returns 403. Zones must be created in the control panel, not via API, with a token like this.

`FindSnapshotRecords` declares no schema-level required inputs on purpose: `snapshotId` feeds the component's own dynamic outPort source, and the e2e rule `dynamic-source-static-input` rejects a required source input bound to a modifier. It is still enforced at run time with a `CancelError`. Flagged for a second opinion.

Bright Data publishes no single hard rate limit; `quota.js` uses a conservative 10 req/s sliding window. Initial release, so `bundle.json` stays at 1.0.0.


vtalas (commented):

**First, the good news:** I reproduced the validator, lint and e2e assertion counts independently and **they match your PR description exactly** — validator 0 failures / 0 warnings, eslint clean, 12 e2e assertions, and the `^s_` → `^sd_` regex fix is present in the flow. Nothing here disputes your test evidence; these are findings the 25 static rules don't express.

**1 error, 5 warnings, 6 info.** Inline comments cover the errors and warnings. Info items are listed below.

### Info (6)

- **`statusCode` vs `status` in MakeApiCall** — `MakeApiCall.js:60` emits `statusCode`; 155 of 158 sibling `MakeApiCall` components emit `status` (only `fal` and `gladia` differ). The connector is at least internally consistent since `ScrapeUrl` does the same, but it's the ecosystem outlier. Cheap to change now — `bundle.json` is 1.0.0 with no released users.
- **Redirects aren't re-pinned** — `lib.resolveApiUrl` validates only the *initial* URL and `MakeApiCall.js:57` sets no `maxRedirects`. Residual risk is low (the redirect has to originate from Bright Data itself, and axios' `follow-redirects` strips `Authorization` cross-host) and **no connector in the repo sets `maxRedirects`**, so this is defense-in-depth, not a bypass.
- **Account-specific e2e fixtures** — `test-flow-zones-scrape-search.json:173` and `:275` hardcode `web_unlocker1` and `serp_api1`, which exist only because they were hand-provisioned on one account. Any other CI account fails both flows with a 400. Feeding the zone from the ListZones step already in the flow would fix it.
- **Engine-specific params silently dropped** — `lib.js:28-36`: `duckduckgo` defines no `count`/`offset` and `yandex` defines none of `country`/`language`/`count`/`offset`. The tooltips caveat only `country` and `numResults` with "Supported by Google and Bing"; `page` and `language` claim to work unconditionally.
- **`body` sent on GET** — `ScrapeUrl.js:31-33` sets `extra.body` whenever `body` is non-empty regardless of `httpMethod`, while `component.json:113` says "Only used when the HTTP method is POST".
- **Silent shape fallback in ListZones** — `ListZones.js:28` turns any unexpected payload (a shape change, or an `{error}` body that still returned 2xx) into "you have no zones" with no log. `FindSnapshotRecords.js:48-53` handles the same situation the opposite way and throws, so the connector is inconsistent with itself.

---

vtalas (commented):
## Follow-up: `TriggerDatasetCollection` is an async component

Second pass on this PR, prompted by @vtalas spotting that the dataset chain matches `14-async-components.md` (*"Async Components — jobs that finish later"*). It does, and **my first review missed it** — see the note at the bottom on why, because that part is a bug in the review skill rather than in this PR.

### Why this is the async shape

`POST /datasets/v3/trigger` returns a `snapshot_id` and the job finishes later. That is the definition. The reference file's decision rule turns on one question — *does the provider offer a callback URL parameter?* — and **Bright Data does**: it is the very `endpoint` parameter this component already sends on `TriggerDatasetCollection.js:46`.

So the prescribed shape here is **self-callback**:

```json
{ "webhook": true,
  "outPorts": [ { "name": "out",  "schema": { "…": "snapshotId + echo" } },
                { "name": "done", "schema": { "…": "records + the same echo" } } ] }
```
```js
params.endpoint = context.getWebhookUrl();
```

What the connector does instead:

| | Prescribed | Current |
|---|---|---|
| `"webhook": true` | required | absent |
| `context.getWebhookUrl()` | required | never called |
| completion port | required | only `out` |
| callback URL | supplied by the component | **exposed as the `notifyUrl` input** |
| how the user gets records | automatically, on the completion port | hand-built `Wait` → `GetSnapshotStatus` → `FindSnapshotRecords` |

That last row is visible in your own e2e flow — `test-flow-dataset-snapshot.json` runs `TriggerDatasetCollection → utils.timers.Wait → GetSnapshotStatus → FindSnapshotRecords`. The PR description notes the `Wait` was tuned down from 5 min to 1 min because "measured readiness is a few seconds", which is exactly the fragility the async pattern exists to remove: a fixed timer that is too short reads a snapshot that isn't ready, and one that is long enough wastes wall-clock on every run.

The four reference implementations are all actions with `webhook: true` and a completion port, so this is an established shape rather than a theoretical one: `clearbit/enrichment/FindPerson` (`person`), `plivo/sms/SendSMSAndWaitForReply` (`response`/`noResponse`), `twilio/calls/ForwardCall` (`completed`), `utils/tasks/RequestApproval` (`approved`).

### Relationship to the `api.param-semantics` warning above

They're the same defect seen from two distances. That comment argues about whether `notify` should be a URL or a boolean — but under rule 14 **neither is right, because it should not be a user input at all.** If you restructure to self-callback, the `notify`/`endpoint` ambiguity stops mattering: the component supplies its own URL and you find out the correct wire format once, at implementation time, instead of documenting a footgun. Treat this comment as superseding that one.

### Scope, honestly

This is a ports change, so it is **breaking after release** — which is the argument for doing it now while `bundle.json` is 1.0.0 with no users, not the argument for rushing it into this PR. Splitting it into a follow-up is completely reasonable; I've filed it as an issue so it doesn't get lost. What I would not do is ship `notifyUrl` as-is, because that input is the one piece the reference file calls out by name as a footgun, and removing it later is also breaking.

### Why the review missed it the first time

Not a judgment call that went the wrong way — `14-async-components.md` ships in the `review-connector` skill's `references/` directory but **was not listed in its SKILL.md reference table, and the word "async" appeared nowhere in the skill**. The only pointer to it was a cross-reference buried inside `07-component-types.md:345`. The same file was orphaned in `build-connector` too, which is the more interesting half: the build side never had the rule either, so the pipeline could neither produce this shape nor catch its absence. Fixes to all three skills are in flight.

vtalas (approved):

PR #1232, opened 2026-08-26
telegram: new connector (11 actions, 2 triggers)

petrolivka:
## Telegram connector (new)

Telegram Bot API connector — 11 actions + 2 triggers. Built and verified end to end against a live bot (`@appmixer_test_bot`) on `dev-automated-00001`.

### Components

| Component | Bot API method | Notes |
|---|---|---|
| SendMessage | `sendMessage` | parse mode, silent, protect, thread, reply |
| SendPhoto | `sendPhoto` | URL / Telegram file ID, or upload a stored file |
| SendDocument | `sendDocument` | as above; 50 MB bot upload cap enforced |
| EditMessage | `editMessageText` | returns the edited message |
| DeleteMessage | `deleteMessage` | returns `{}` |
| ForwardMessage | `forwardMessage` | |
| GetChat | `getChat` | |
| GetFile | `getFile` + download | stores the file for later steps |
| AnswerCallbackQuery | `answerCallbackQuery` | pairs with the callback trigger |
| ListChatAdministrators | `getChatAdministrators` | `outputType` via lib helpers |
| MakeApiCall | any | origin-locked to `api.telegram.org` |
| **NewMessage** *(trigger)* | webhook | chat / edited / channel-post filters |
| **NewCallbackQuery** *(trigger)* | webhook | chat + callback-data filters |

Auth: API key (bot token), validated with `getMe`. Quota: 25 req/s per user, under Telegram's ~30/s ceiling.

### Why the plugin webhook pattern

Telegram allows **exactly one webhook per bot** and refuses `getUpdates` while one is registered, so the per-trigger `context.getWebhookUrl()` model cannot work — a second trigger would silently overwrite the first one's registration. This uses `plugin.js` + `routes.js` (same shape as the LINE connector): every trigger of a bot registers the same derived URL and subscribes via `addListener`, and one route fans updates out to matching listeners.

- Path segment is `sha256(botToken + ':webhook-id')`, so **no credential appears in a URL**.
- Authenticity is enforced in the `triggerListeners` filter: Telegram's `X-Telegram-Bot-Api-Secret-Token` header is compared against the value the listener registered (derived from a *different* input, so the public path never reveals it). A forged POST to a guessed path matches no listener and is dropped.
- `stop()` removes only the listener, never the webhook — the registration is shared with sibling flows.

### Test evidence

**Static** — `eslint` exit 0; `node scripts/validate.js --connector telegram`: **0 failures, 0 warnings** (25 validators); `appmixer e2e validate`: **passed** (26 validators, 5 flows).

**CLI component tests** (live bot, group `Appmixer Test`) — 10 of the 13 components run, **10 pass**: MakeApiCall, SendMessage, EditMessage, ForwardMessage, SendPhoto, SendDocument, GetChat, ListChatAdministrators, GetFile, DeleteMessage. Error normalization and required-input assertion checked explicitly (`getChat` on a bogus id → `Telegram getChat failed. Bad request (400). Bad Request: chat not found`).

**E2E** (`dev-automated-00001`) — **all 5 flows verified in the E2E success store**:

| Flow | Result |
|---|---|
| chat info | ✅ PASSED |
| messaging | ✅ PASSED |
| media (incl. **multipart upload** via SendDocument) | ✅ PASSED |
| new message trigger | ✅ PASSED — plugin webhook verified end to end |
| callback query trigger | ✅ PASSED — real inline-keyboard button press |

The webhook registered at exactly the derived URL, and the path hash matches `lib.getWebhookId` locally. A real bot message delivered to the plugin endpoint inside a genuine update envelope fired the trigger with the correct secret; the wrong-secret POST produced no second delivery. The callback flow was driven by an actual button press, which fired `NewCallbackQuery` and then `AnswerCallbackQuery`.

### Notes for reviewers

- Both triggers are verified at runtime. The two trigger flows need **one manual action each** to reproduce (reply to a message / press a button), since Telegram never delivers a bot its own messages — steps are in `artifacts/test-flows/README.md`.
- One connector defect was found by live testing and fixed: `GetFile` passed an axios stream to `saveFileStream`, which takes a Buffer (`fcad6b96`) — now `arraybuffer`, matching `box/DownloadFile`.
- `SendPhoto`/`SendDocument` multipart upload is **not** reachable from `appmixer test component` (its file store is per-invocation), which is why the media E2E flow deliberately round-trips GetFile → SendDocument.
- `package.json` adds `form-data@4.0.0` for multipart uploads, following `box`.
- E2E flows hard-code the test chat id `-5467341409`; see the README to retarget.
- Gotcha worth knowing for other E2E flows: `Assert` renders a modifier binding into the clause as a string, so a JSON boolean expectation never matches and fails with the confusing `expected true to equal true`. Write `"true"` / `"false"` as strings (`c4639839`).

### Review follow-up (2026-08-27)

All 8 findings from the `review-connector` pass verified against the code and addressed in `eef067f5` + `c3ace705`:

| Finding | Done |
|---|---|
| stray `file_0.jpg` committed in GetFile/ | deleted, and the pattern the CLI writes is now gitignored |
| dynamic outPort bound `in/chatId` to the flow value | now the `"dummy"` literal, matching 12 other connectors |
| flows carry no `notes` sticky | all 5 flows now state the required chat, bot membership and the manual step |
| ForwardMessage lacks `forward_origin` | added, with `is_automatic_forward` |
| one Message schema copy-pasted across 5 components | trimmed to what each method actually returns |
| SendPhoto guarded against the 50 MB document limit | `lib.sendMedia` now uses a per-method ceiling (10 MB / 50 MB) |
| `pipeline-state.json` inaccuracies (2×) | corrected, here too |

One deliberate deviation: `edit_date` is **kept** on EditMessage. `editMessageText` does stamp it on the returned message, and the messaging E2E flow asserts `notEmpty` on exactly that field — dropping it would have broken a passing flow.

Re-published and re-ran the three unattended flows against the live bot after the changes: chat info, messaging and media all ✅ in the E2E success store (12:50–12:51 UTC). `validate.js --connector telegram` stays at 0/0 and `appmixer e2e validate` still passes.

https://claude.ai/code/session_01U6tC3rNLycwXHzK256ro1E


vtalas (commented):
## 🤖 Connector review

**0 error(s), 5 warning(s), 3 info** across 7 component(s).

This PR adds a new Telegram Bot API connector: 11 actions, 2 webhook triggers, apiKey auth on a bot token, a shared lib.js, a quota manager, and a plugin route that fans Telegram's single per-bot webhook out to listeners. I reviewed all 13 components (component.json + behavior) against the review-connector rule set, plus auth.js, lib.js, routes.js, quota.js, service.json, bundle.json and the five E2E flows, and read the `appmixer connector validate --changed` output (all 25 validators pass, 1 warning). The architecture is unusually careful — the webhook id and setWebhook secret_token are derived from separate hashes so the bot token never reaches a URL, the shared registration is deliberately not deleted on stop, MakeApiCall has an explicit origin check to stop the token being used as an SSRF/exfiltration primitive, and every trigger's test() reuses the same update-to-payload mapping as the live path. No errors found. The findings are a stray committed binary, the one validator warning, and over-declared output schemas copy-pasted across the message-returning components.

|---|---|---|---|---|
| ⚠️ warning | appmixer.telegram.core.GetFile<br><sub>src/appmixer/telegram/core/GetFile/file_0.jpg:1</sub> | `repo.stray-artifact` | A 3.8 KB binary `file_0.jpg` — a leftover from live-testing the download path — is committed inside the GetFile component directory, where the platform will treat it as part of the component bundle. | Delete `src/appmixer/telegram/core/GetFile/file_0.jpg` from the branch. |
| ⚠️ warning | appmixer.telegram.core.ListChatAdministrators<br><sub>src/appmixer/telegram/core/ListChatAdministrators/component.json:81</sub> | `validate.dynamic-outport-required-inputs` | The dynamic outPort source maps the required input `in/chatId` from the flow value `inputs/in/chatId`, so when a user binds Chat ID to a variable the designer's required-input validation rejects the schema call and the variable picker silently degrades to Raw Output. | Send the literal `"dummy"` instead — `"in/chatId": "dummy"` — since chatId does not shape the generated schema. |
| ⚠️ warning | appmixer.telegram.core.SendMessage<br><sub>src/appmixer/telegram/core/SendMessage/component.json:245</sub> | `schema.invented-fields` | One Message out-schema is copy-pasted across SendMessage, EditMessage, SendPhoto, SendDocument and ForwardMessage, so each declares fields its own Bot API method can never return — `photo`/`document`/`caption`/`edit_date` on sendMessage and editMessageText, `photo`/`text`/`edit_date` on sendDocument, `edit_date`/`reply_to_message` on forwardMessage — which fills the variable picker with entries that are always empty. | Trim each component's out-schema to the fields its method actually returns (e.g. drop photo/document/caption/edit_date from SendMessage and EditMessage, drop photo/text/edit_date from SendDocument), keeping the full Message shape only where it genuinely applies such as NewMessage. |
| ⚠️ warning | appmixer.telegram.core.ForwardMessage<br><sub>src/appmixer/telegram/core/ForwardMessage/component.json:95</sub> | `schema.missing-key-field` | Because the out-schema is the shared Message copy, it omits `forward_origin` (and `is_automatic_forward`) — the field identifying the original sender, which is the main reason a flow reads a forwardMessage result. | Add a `forward_origin` object with type/title/example to the ForwardMessage out-schema. |
| ⚠️ warning | appmixer.telegram.core.NewMessage<br><sub>src/appmixer/telegram/artifacts/test-flows/test-flow-newmessage-trigger.json:1</sub> | `e2e.missing-data-assumption-note` | All five E2E flows hardcode the tenant-specific chat ID `-5467341409` and the two trigger flows additionally need a human to send a message or press an inline button, yet no flow carries the top-level `notes` sticky that rule 19 requires — imported on a fresh tenant they just time out after 600s with no explanation in the designer. | Add a top-level `notes` entry to each flow stating which chat must exist, that the bot must be a member, and the manual step needed for the two trigger flows. |
| ℹ️ info | appmixer.telegram.core.SendPhoto<br><sub>src/appmixer/telegram/lib.js:229</sub> | `validation.upload-limit-mismatch` | SendPhoto's own description states Telegram caps photo uploads at 10 MB, but lib.sendMedia guards every upload against the 50 MB document limit, so an 11–50 MB photo passes the local check and comes back as an opaque "Bad request (400)" from Telegram. | Make the size ceiling a per-method parameter of sendMedia — 10 MB for sendPhoto, 50 MB for sendDocument — so the CancelError names the real limit. |
| ℹ️ info | appmixer.telegram.core.SendPhoto<br><sub>src/appmixer/telegram/artifacts/ai-artifacts/pipeline-state.json:47</sub> | `docs.inaccurate-artifact` | pipeline-state.json records the SendPhoto multipart-upload path as "covered by the media E2E flow instead", but that flow drives SendPhoto with a plain URL (test-flow-media.json:41) and only SendDocument exercises the upload branch, so SendPhoto's multipart path is verified by neither the CLI runs nor E2E. | Either feed GetFile's stored fileId into SendPhoto.fileId in the media flow, or correct the artifact to record the multipart path as covered for SendDocument only. |
| ℹ️ info | appmixer.telegram.core.MakeApiCall<br><sub>src/appmixer/telegram/artifacts/ai-artifacts/pipeline-state.json:33</sub> | `docs.inaccurate-artifact` | pipeline-state.json claims "11 of 13 components tested via appmixer test component. 11 pass", but the adjacent `passed` array lists only 10 entries, which together with the 3 components marked not-testable-via-CLI accounts for all 13. | Correct the count to 10 so the verification record matches its own evidence list. |

_7 of these are also posted inline on the changed lines._

- All 25 validators in `appmixer connector validate --changed` pass; no failures to convert into findings
- Bot token never placed in a webhook URL — webhookId is sha256(token+':webhook-id'), the setWebhook secret_token is a separate hash, and routes.js authenticates by comparing the echoed X-Telegram-Bot-Api-Secret-Token in the listener filter
- MakeApiCall rejects any absolute URL whose origin is not https://api.telegram.org before appending the bot token, closing an authenticated SSRF / token-leak path
- Both triggers implement test(context): read-only via getUpdates, write no state, honour context.properties filters, emit exactly one item, and throw a CancelError (never fabricate data) when no real sample exists
- Both triggers share their update-to-payload mapping with the live webhook path via lib.messageFromUpdate / lib.callbackFromUpdate, so Flow Test Mode cannot drift from a real run
- stop() removes only the listener and deliberately does not call deleteWebhook, so one flow stopping cannot silence sibling flows sharing the bot's single registration
- ListChatAdministrators uses the required lib.sendArrayOutput / lib.getOutputPortOptions helpers, the `result` array field, all four outputType modes, and a single shared schema const with type+title+example on the leaves
- DeleteMessage returns `context.sendJson({}, 'out')` per the delete/update rule; AnswerCallbackQuery correctly normalises Telegram's bare `true` into `{ success }`
- Every action asserts its required inputs with context.CancelError before calling the API, and all API failures route through lib.normalizeError, which surfaces Telegram's own description plus retry_after on 429
- inPort JSON Schema and inspector.inputs are in exact 1:1 parity across all 13 components; no `select` carries a `source`; integer fields use `number` inspectors and booleans use `toggle`
- All 13 components carry the quota block referencing manager appmixer:telegram / resources `requests`, which quota.js defines at 25 requests/second under Telegram's ~30/s ceiling
- No hardcoded credentials anywhere — the only token-shaped string is Telegram's own documentation placeholder in the auth.js tooltip
- No prompt-injection or agent-directed instructions found in any connector source, README, test fixture or artifact
- All 13 components are exercised by at least one E2E flow, and all 15 assertions use only `equal` / `notEmpty` with the correct `expected` key and booleans quoted as strings

</details>

petrolivka:
Thanks — I verified all 8 findings against the code before acting on them and every one held up, including the line references. All are addressed in `eef067f5` and `c3ace705`.

| Finding | What I did |
|---|---|
| `repo.stray-artifact` | Deleted `file_0.jpg`. It was a leftover from the live GetFile run that got swept into `fcad6b96`. Also added `src/appmixer/*/core/*/file_*` to `.gitignore`, since that is the exact shape the CLI writes into the component directory. |
| `validate.dynamic-outport-required-inputs` | Now `"in/chatId": "dummy"`. You were right that this was the wrong way to silence the validator — I had added the real binding to clear the failure without thinking through what it does to the designer. `beehiiv/FindPosts` and 11 other connectors use the literal, and the validator still reports 0 issues with it. |
| `schema.invented-fields` | Each of the five out-schemas trimmed to what its own method returns — see the deviation note below. |
| `schema.missing-key-field` | ForwardMessage gained `forward_origin` (MessageOrigin: `type`, `date`, `sender_user`, `sender_user_name`, `sender_chat`, `chat`, `message_id`, `author_signature`) plus `is_automatic_forward`. |
| `e2e.missing-data-assumption-note` | All five flows now carry a top-level `notes` sticky naming the chat id, the bot's membership requirement and how to retarget; the two trigger flows additionally carry a `MANUAL STEP REQUIRED` block spelling out the reply / button press and the 600s timeout. |
| `validation.upload-limit-mismatch` | `lib.sendMedia` guards against a per-method ceiling. I keyed it by method (`UPLOAD_LIMITS = { sendPhoto: 10MB, sendDocument: 50MB }`) rather than adding a parameter, so a caller cannot pair a method with the wrong ceiling. An oversized photo now fails with `Maximum upload size for sendPhoto is 10MB`. |
| `docs.inaccurate-artifact` ×2 | Both corrected. The CLI run covered 10 components, not 11, and the multipart branch is exercised by SendDocument only — the artifact now says so, and so does the PR description, which carried the same "11 pass" claim. |

**One deliberate deviation.** I kept `edit_date` on EditMessage rather than dropping it: `editMessageText` does stamp it on the message it returns. `test-flow-messaging.json` asserts `notEmpty` on exactly `EditMessage.out.edit_date` and has been passing, so following the suggestion literally would have broken a green flow. Dropped it from SendMessage, SendPhoto, SendDocument and ForwardMessage as you described.

**Re-verified after the changes.** `lib.sendMedia` is behaviour, so I re-published and re-ran the three unattended flows against the live bot — chat info, messaging and media are all ✓ in the E2E success store (12:50–12:51 UTC today). `validate.js --connector telegram` stays at 0 failures / 0 warnings and `appmixer e2e validate` still passes. The two trigger flows need a human action and their schemas did not change, so they still carry yesterday's green result; happy to re-provoke them if you would rather see them re-stamped.

petrolivka:
- https://dev-automated-00001.dev.appmixer.ai/designer/30afaa58-032c-4a55-8cdd-b5fe8b977348
- https://dev-automated-00001.dev.appmixer.ai/designer/f04fa3b5-cae4-45e0-ba3b-287cc037d6da
- https://dev-automated-00001.dev.appmixer.ai/designer/07278919-0036-46c5-8feb-a2e30f61e9aa
- https://dev-automated-00001.dev.appmixer.ai/designer/32dea996-7ae2-4ff5-9449-de8cf81a8407
- https://dev-automated-00001.dev.appmixer.ai/designer/21cd62b2-a039-46d4-bd98-dc484a880fdc

apx-vero:
E2E report for this PR: Appmixer-ai/appmixer-components#2383 (comment https://github.com/Appmixer-ai/appmixer-components/issues/2383#issuecomment-5454531162).

**5/5 flows pass.** The three automatic ones (chat info, messaging, media) were re-run today against a fresh `remove` + `publish` of this branch; `validate.js --connector telegram` is 0/0 and eslint is clean. The two trigger flows still carry their pass from two days ago — they need a human in the Appmixer Test group (reply to the bot's prompt / press **Fire E2E**), so I could not re-run them unattended.

This PR has no issue reference in its body and I don't have write access to edit it — worth adding `Closes Appmixer-ai/appmixer-components#2383` so the link is explicit.

vtalas (approved):

PR #1235, opened 2026-08-26
feat(ci): link the maya review run and draft the PR on error findings

vtalas:
Follow-up to #1234. Three additions to the `apx-maya` connector review workflow.

### 1. The run statistics stop depending on a hardcoded path

`Run summary` read `$RUNNER_TEMP/claude-execution-output.json` directly — the one
way the cost table could silently degrade to *"No Claude execution output found"*.
The path now comes from the action's own `execution_file` output, with that
RUNNER_TEMP path as the fallback (the uploaded artifact follows the same rule).

Parsing moved into a new `stats` step right after Claude so the review comment can
quote the numbers; the table still renders at the end of the job summary, and gains
cached input tokens (without them the input count reads misleadingly small), the PR
link, and the draft flip.

| Metric | Value |
|---|---|
| Result | success |
| Duration | 3m 13s |
| Turns | 37 |
| Input tokens | 812 |
| Cached input tokens | 1,932,244 |
| Output tokens | 15,234 |
| Cost (USD) | $1.2346 |

### 2. The review links back to its run

New footer line on the review body — and on the fallback comment when the findings
file is unreadable:

> <sub>[Run details](https://github.com/Appmixer-ai/appmixer-connectors/actions) · 3m 13s · $1.2346</sub>

Duration and cost are repeated inline so the common case needs no click; the link
leads to the full table above.

### 3. Error findings put the PR back into draft

A review that finds `error`-level issues converts the PR to draft via
`gh pr ready --undo`, and the review body says why:

> [!IMPORTANT]
> This PR was moved back to **draft** because the review found 1 error-level finding(s).
> Address them and press **Ready for review** — then re-request a review from `apx-maya` to run this again.

Controlled by a workflow-level knob:

```yaml
DRAFT_ON: error          # error | any | never
```

`error` by default on purpose — a warning-only review should not pull a PR out of
the review queue. Never the reverse: marking a PR ready again stays the author's
call. The step runs *before* the review is posted (so the body can explain the
state change), skips PRs that are already drafts, and is non-fatal — if the flip
fails the review is still posted, with a `::warning::` in the log.

### ⚠️ Needs a permission change

`apx-maya` needs **write** access on this repo for the draft flip. Every other step
only ever needed read. Without it the review still posts and the draft flip logs a
warning.

### Testing

No runner needed for any of this — the embedded scripts were extracted and exercised
locally: the stats parser against a fixture execution log, the review-body builder
(table, inline-anchor selection, draft callout, footer), the draft logic across all
branches (`error`/`any`/`never`, missing findings file, PR already draft, `gh`
failure), and the summary renderer with and without stats. YAML parses.

Worth a live `workflow_dispatch` against a PR with a known error finding before
relying on the draft flip.

https://claude.ai/code/session_01EBPSSwQSs7w98vxsfNu8zq

### DIFF
--- a/.github/workflows/claude-maya-review.yml
+++ b/.github/workflows/claude-maya-review.yml
@@ -50,7 +50,16 @@ name: Claude Maya Review (connector code review)
 # Re-requesting a review from apx-maya re-runs the whole thing (posting a
 # review clears the request, so this is the natural "review again" gesture).
 #
-# Testing: use "Run workflow" (workflow_dispatch) with a PR number.
+# RUNNING IT EXPLICITLY: "Run workflow" (workflow_dispatch) with a PR number
+# reviews any open PR, from `dev`, regardless of what the PR's own branch
+# contains — the workflow definition comes from the dispatch ref, and the review
+# checks out refs/pull/<n>/head. A manual run is SELF-AUTHORIZING: dispatching
+# already requires write access on this repo, exactly the permission needed to
+# request a review from apx-maya, so it does not additionally require the review
+# request and it bypasses the iteration cap. It is the way out when the
+# re-request gesture is unavailable (e.g. apx-maya is still listed as a
+# requested reviewer because a previous review was posted under a different
+# identity, so the request was never cleared).
 #
 # Required secrets: ANTHROPIC_API_KEY, MAYA_GH_TOKEN (PAT of the apx-maya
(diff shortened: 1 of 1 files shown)


vtalas:
Superseded by #1236 — same branch, opened under the apx-vero account.

PR #1240, opened 2026-08-27
gohighlevel(patch): add Ed25519 X-GHL-Signature webhook verification helper

apx-vero:
## Context

HighLevel is deprecating the legacy webhook signature header `X-WH-Signature` (RSA-SHA256) on **September 1, 2026**. After that date webhook requests are signed only with Ed25519 and the signature is delivered in `X-GHL-Signature`.

Docs: https://marketplace.gohighlevel.com/docs/webhook/WebhookIntegrationGuide

## Finding

There is **nothing to migrate**. `src/appmixer/gohighlevel/` ships no trigger components at all (only Create/Get/Find/List/Update/Delete actions plus `MakeApiCall`), and a repo-wide search for `X-WH-Signature` / `signature` in the connector returns no hits. No Appmixer code currently consumes the deprecated header, so the deprecation causes no breakage today.

## Change

Rather than leave the next GoHighLevel trigger to rediscover this, the supported verification scheme is now available in the connector's shared lib:

- `src/appmixer/gohighlevel/lib.js`
  - `WEBHOOK_PUBLIC_KEY` — HighLevel's published Ed25519 public key (PEM), with a comment recording the deprecation and a link to the docs.
  - `lib.WEBHOOK_SIGNATURE_HEADER` — `'x-ghl-signature'`.
  - `lib.verifyWebhookSignature({ rawBody, signature })` — verifies the base64 signature over the **raw** request body via `crypto.verify(null, ...)`. Returns `true` only on a valid signature (also `false` on a missing/`'N/A'` signature or a malformed key/signature), so callers reject on `false`.
- `src/appmixer/gohighlevel/bundle.json` — version `1.1.0` → `1.1.1` with a changelog entry.

The helper is intentionally the whole change: it is a small, self-contained building block, and adding a webhook trigger is a separate piece of product work that is out of scope for this deprecation notice. If reviewers would rather not carry a currently-uncalled helper, closing this and tracking the deprecation against a future trigger is a reasonable alternative — flagging that explicitly rather than burying it.

## Verification

- `npm run lint` — clean.
- `node scripts/validate.js --changed --base origin/dev` — all 25 validators OK.
- `node scripts/validate.js --connector gohighlevel` reports 81 findings, **all pre-existing** in `component.json` files this PR does not touch (`output-port-examples` ×77, `no-select-with-source` ×2, `required-inputs-asserted` ×2). Not addressed here to keep the diff scoped to the deprecation.
- `appmixer connector verify --offline` skipped: `src/appmixer/gohighlevel/artifacts/samples/` does not exist.
- No test covers the helper — the repo has no unit tests for connector libs, and this step had no access to a live instance.

## Note on the source issue

The issue body is a pasted vendor announcement email containing a tracking/redirect link (`email.mailbox.gohighlevel.com/c/...`). It was not followed; the public-key and verification details above come from the canonical marketplace docs URL. The issue contained no instructions beyond the deprecation notice.

Closes Appmixer-ai/appmixer-components#2806

### DIFF
--- a/src/appmixer/gohighlevel/artifacts/test-flows/test-flow-appointment-scheduling.json
+++ b/src/appmixer/gohighlevel/artifacts/test-flows/test-flow-appointment-scheduling.json
@@ -1,64 +1,87 @@
 {
     "flow": {
-        "after-all": {
+        "0d18d15f-3fca-5e41-b2d3-6e21ec237000": {
             "config": {
                 "properties": {
                     "timeout": 30
                 }
             },
+            "errorHandling": {
+                "autoRetry": false,
+                "onError": "stopFlow"
+            },
             "source": {
                 "in": {
-                    "assert-appointment": [
+                    "4e3ecc6f-88b8-55c9-ad97-f321a55cfa09": [
                         "out"
                     ],
-                    "assert-list-calendars": [
+                    "caeca176-d348-5aea-8dd5-7d2e8751434d": [
                         "out"
                     ]
                 }
             },
             "type": "appmixer.utils.test.AfterAll",
-            "x": 1360,
-            "y": 80
+            "x": 1184,
+            "y": 16
         },
-        "assert-appointment": {
+        "33802334-31cf-5169-9a25-2f2a579151b0": {
             "config": {
+                "properties": {
+                    "account": "69bd20b72e33e81a1e686f91"
+                },
                 "transform": {
                     "in": {
-                        "create-appointment": {
+                        "afd3b15e-b99c-5056-afb5-fb5f48af6d1a": {
                             "out": {
                                 "lambda": {
-                                    "expression": {
-                                        "AND": [
-                                            {
-                                                "assertion": "notEmpty",
-                                                "field": "{{{appt-id-var}}}"
-                                            },
-                                            {
-                                                "assertion": "equal",
-                                                "expected": "E2E Test Appointment",
-                                                "field": "{{{appt-title-var}}}"
-                                            },
-                                            {
-                                                "assertion": "equal",
-                                                "expected": "new",
-                                                "field": "{{{appt-status-var}}}"
-                                            }
-                                        ]
-                                    }
+                                    "address": "123 E2E Street, Austin, TX",
+                                    "appointmentStatus": "new",
+                                    "calendarId": "{{{var-calendar-id}}}",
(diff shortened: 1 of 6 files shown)


apx-vero:
Published + imported E2E flow(s) (NOT started):

- [E2E Gohighlevel - Contact CRUD](https://dev-automated-00001.dev.appmixer.ai/designer/4a79eb44-c886-4648-871b-f3eec65e977c)
- [E2E Gohighlevel - Find Contacts](https://dev-automated-00001.dev.appmixer.ai/designer/b3e870a9-8a0c-4661-959b-165c0589a8cf)
- [E2E Gohighlevel - Opportunity Lifecycle](https://dev-automated-00001.dev.appmixer.ai/designer/7c7a0483-d7db-4eff-b687-048ca0f29583)
- [E2E Gohighlevel - Appointment Scheduling](https://dev-automated-00001.dev.appmixer.ai/designer/f38c4f77-f951-4662-ae8d-7d702b66c048)

`appmixer e2e validate` passes on all four (27 validators) and `appmixer e2e import -c gohighlevel` exited 0 — accounts bound, no red-chip variables, no account-validity failures. All four flows are in stage `stopped`.

Every component was already on the instance at 1.0.0, so a plain `publish` would not have refreshed the runtime bundle — and this PR's only code change is in the shared `lib.js`. I removed `appmixer.gohighlevel.core` and republished so the instance actually runs the code under review.

Flow changes in a2bdf8ef cover the two hard validator failures plus the warnings that pointed at real gaps: UUID component ids, fail-fast `errorHandling` on every node, MakeApiCall coverage doubling as appointment cleanup via `DELETE /calendars/events/{eventId}`, an assert on the previously unchecked CreateContact output, two CodeBlocks replaced by `g_addTimeSpan`/`g_timestamp` modifiers, unique per-run emails, and a grid re-layout.

One caveat worth knowing before the first run: the cleanup node passes the required `Version: 2021-07-28` header through MakeApiCall's key-value `headers` input as a JSON string — the documented form, and what other connectors' flows do. But the connector's `MakeApiCall.js` only reads that input when it arrives as an array, so this relies on engine-side coercion that I could not verify: there is no local gohighlevel auth for a component test, and I did not start any flow. If the first run fails on the cleanup node, that header is the thing to look at.

vtalas (approved):

PR #1241, opened 2026-08-28
requesty (major): align with the AI connector contract, add GenerateEmbeddingsFromFile

apx-vero:
## Summary

Brings `ai/requesty` to the AI connector contract proposed in Appmixer-ai/appmixer-components#2808, and adds the missing `GenerateEmbeddingsFromFile`.

The connector is not on the release branch, so it can take the renames now instead of carrying them as a migration later. That also makes it the reference implementation for the contract work: it is the only AI connector at **0 validator failures, 0 warnings**.

## What changed

**Renames** — the same operation currently ships under four names across the eight AI connectors; these are the two outliers:

| before | after | matches |
|---|---|---|
| `ChatCompletion` | `SendPrompt` | claude, gemini, openai, openrouter, groq, bedrock |
| `CreateEmbedding` | `GenerateEmbeddings` | openai, gemini, voyageai |

**Output shapes**, so a flow can swap providers without rewiring every downstream binding:

- `SendPrompt` emits `answer` + `prompt` alongside `id` / `model` / `created` / `finishReason` / `usage.*`. It previously collapsed `choices` to its first element and exposed the raw `choices.message.content` path, which no other connector uses.
- `GenerateEmbeddings` emits `embeddings[]{index, text, vector}` + `firstVector`, the shape openai/gemini/voyageai use and the one `pinecone.InsertVectors` consumes without a modifier. It previously returned a single flat vector. `input` became `text`, and `encodingFormat` was replaced by `chunkSize` / `chunkOverlap` (500 / 50), matching the other embedding components.

**New `GenerateEmbeddingsFromFile`**, mirroring openai/gemini: the file is read in 1MB parts, each part is chunked and embedded, and one message is emitted per part.

**`MakeApiCall` pins the origin.** It attaches the connector's API key to whatever URL it is given, so an absolute URL on another host sent the credential there — Appmixer-ai/appmixer-components#2802. `lib.resolveApiUrl` now rejects anything off `https://router.requesty.ai`, and URLs carrying embedded credentials.

**Input order** on `SendPrompt` is now `prompt`, `instructions`, `model`, then the Advanced group — the order the contract proposes.

**`lib.js`** gains `resolveApiUrl`, a `request()` wrapper and the shared embedding helpers (`splitText`, `generateEmbeddings`); `ListModels` uses the wrapper instead of building its own request.

## Testing

- `node scripts/validate.js --connector ai/requesty` → **0 failures, 0 warnings**
- `npx eslint src/appmixer/ai/requesty --ext .js` → clean
- Behaviour smoke-tested against a stubbed `context`: required-input assertions fire for all four components; `GenerateEmbeddings` batches correctly (18 chunks from an 800-word input at `chunkSize: 100`); `GenerateEmbeddingsFromFile` streams and emits per part.
- Origin pinning verified: `/v1/models`, `v1/models` and the full `https://router.requesty.ai/...` URL resolve; `https://evil.example.com/steal`, `//evil.example.com/steal` and `https://user:pass@router.requesty.ai/...` all raise `CancelError`.

E2E flows follow the renames (`test-flow-send-prompt`, `test-flow-generate-embeddings`) with assertions moved onto the new output fields, and a new `test-flow-generate-embeddings-from-file` covers the added component (SaveFile → embed → Assert → AfterAll → RemoveFile). **Not yet run against a live instance** — worth doing before merge.

## Version

`2.0.0`. The renames are breaking for anyone who installed 1.0.0 from `dev`, even though the connector never reached the release branch.

## Related

Part of the AI connector consistency work: Appmixer-ai/appmixer-components#2808 (contract) and the per-connector issues #2809–#2816. This is the only one of the eight that could be fixed in place rather than filed.


apx-vero:
## E2E run — all 5 flows pass on `dev-automated-00001`

| flow | result |
|---|---|
| E2E Requesty - Send Prompt | ✅ PASSED |
| E2E Requesty - Generate Embeddings | ✅ PASSED |
| E2E Requesty - Generate Embeddings From File | ✅ PASSED |
| E2E Requesty - List Models | ✅ PASSED |
| E2E Requesty - Make API Call | ✅ PASSED |

All five passed on the first attempt, 0 deterministic retries, 0 errors.

**Verified the passes are real, not vacuous.** The runner's triage does not always read the Assert `error[]` array, so a flow can report PASSED while its assertions never evaluated. A negative-control copy of the embeddings flow — same pipeline, assertion deliberately inverted — was imported, run and deleted:

```
[ERROR]: [appmixer.utils.test.Assert] Assertion failed:
         expected THIS_MUST_NOT_MATCH to equal [object Object]
RESULT | FAILED | E2E Requesty - NEGATIVE CONTROL
```

It failed, and the error shows the variable resolved to the real embeddings array rather than `undefined` — so the assertion machinery is evaluating live component output.

## Instance state

The connector was republished as part of this: the previously published `appmixer.ai.requesty.ChatCompletion` and `…CreateEmbedding` were removed, and the instance now serves exactly the five components this PR defines (`remove` + `publish`, since `publish` alone does not refresh existing component snapshots).

**An auth account for `appmixer:ai:requesty` had to be created.** The instance only had one for `appmixer:requesty` — the service name from before the connector moved under `ai/` — which means this connector has never been E2E-tested since that move. Worth checking whether the other connectors moved in the same batch have the same gap.

vtalas (approved):

PR #1242, opened 2026-08-28
ai/openai(major): drop ghost input, fix output examples, make AIAgent model bindable

apx-vero:
Closes Appmixer-ai/appmixer-components#2813

Part of the AI connector contract work. `ai/openai` had 28 validator failures — the most of any AI connector — and, being the component source most other AI connectors were copied from, its defects had fanned out. `node scripts/validate.js --connector ai` now reports **zero failures for `ai/openai`**.

## What changed

**1. `embeddingTemplate` deleted** — declared in the inPort schema of `GenerateEmbeddings` and `GenerateEmbeddingsFromFile`, present in no inspector and no behaviour file. Also removed from `gemini/GenerateEmbeddingsFromFile`, which inherited it by copy.

**2. `examples` → `example`** — every outPort option used the JSON Schema plural array, which Appmixer does not read, instead of the singular `example` that drives the variable-picker preview. Converted across all 13 components, with realistic values.

**3. `AIAgent`: `model` and `instructions` moved from `properties` to `inPorts`** so they can be bound to a flow variable, matching `ai/claude` and `ai/gemini`. This is the breaking part, hence the major bump. The behaviour falls back to the old `context.properties` values, so agents configured before 2.0.0 keep working — they just need a re-save to pick up the new inputs.

**4. `SendPrompt`: added `instructions`** — the system message was hardcoded to `'You are a helpful assistant.'` with no way to change it. That string is now the default rather than the only option.

**5. `SendPrompt`: added the five sampling inputs** `temperature`, `maxTokens`, `topP`, `frequencyPenalty`, `presencePenalty`, matching groq and requesty. They live in a collapsed "Advanced" group, and only the ones the user actually sets are sent, so the model's own defaults are otherwise untouched.

**6. Embeddings `index` typed `integer`** — the behaviour always writes `i + index`, a number.

**7. `required` now matches the behaviour** — `["text"]` on `GenerateEmbeddings` and `["fileId"]` on `GenerateEmbeddingsFromFile`, both of which the code already asserts.

**8. `ignoreAuth=true`** added to the `TransformTextToJSON` and `VariableExtractor` dynamic outPort `source.url`. Both build their options before any API call, so no authenticated account is needed.

## Two adjacent defects found while auditing the same files

- **`TransformTextToJSON` crashed on every execution.** The required-input guard read `jsonSchema` four lines above the `let jsonSchema` declaration, so each call died in the temporal dead zone with `Cannot access 'jsonSchema' before initialization`. It now guards on the raw `jsonSchemaString`, which is what was intended.
- **`DescribeImages`'s `model` input was invisible.** Its inspector entry was nested inside the `images` field object rather than alongside it, so the schema declared a `model` input the Designer never rendered — the same ghost-input shape as finding 1. Lifted to a top-level input.

Also flipped `AIAgent`'s `storeId` from `select` to `text` (typeahead), the remaining `no-select-with-source` failure in the connector's 28.

## Not done, deliberately

Issue item 8 also notes that `TransformTextToJSON` binds `in/jsonSchema` to a flow variable, which the validator warns about. For `TransformTextToJSON` and `VariableExtractor` that bound value *is* what the output schema is derived from — replacing it with the literal `"dummy"` the warning suggests would produce an empty picker. Left as-is; these two warnings are the only findings remaining on `ai/openai`.

The `examples` → `example` conversion in gemini, voyageai, claude and openrouter is left to their own issues, as #2813 scopes it to the copy source. The one gemini file touched here (`GenerateEmbeddingsFromFile`) got the full treatment so the branch stays green under `--changed`.

## Versions

- `appmixer.ai.openai` 1.4.0 → **2.0.0** (breaking: `properties` → `inPorts`)
- `appmixer.ai.gemini` 1.1.0 → **1.1.1** (ghost input removal only)

## Verification

- `npm run lint` — passes
- `node scripts/validate.js --connector ai` — zero failures for `ai/openai` (down from 28)
- `appmixer connector verify --offline` — skipped, the connector ships no `artifacts/samples/`
- `npm run test-unit` — **not run**: not on this environment's command allow-list. The existing `SendPrompt.test.js` asserts `data.messages[1].content`, which the unconditional system message keeps in place, and `AIAgent.test.js` only covers `mcpCallTool`/`isMCPserver`, neither of which was touched — but this is reasoning, not a green run, so please confirm in CI.

No E2E flows generated and nothing published; that is phase 2.


apx-vero:
Published + imported E2E flow(s) (NOT started):

| Flow | Designer link |
|---|---|
| E2E AI OpenAI - Prompting | https://dev-automated-00001.dev.appmixer.ai/designer/74fa7ecd-b410-47df-aa7f-4fa0a477982d |
| E2E AI OpenAI - Structured Output | https://dev-automated-00001.dev.appmixer.ai/designer/5454bdf2-174d-4d92-8745-d82254718609 |
| E2E AI OpenAI - Embeddings | https://dev-automated-00001.dev.appmixer.ai/designer/ab5234ba-53f5-4075-a28e-9c8d863491d9 |
| E2E AI OpenAI - Media | https://dev-automated-00001.dev.appmixer.ai/designer/39862e36-4336-4a0f-a29d-1ec8b16f47e0 |
| E2E AI OpenAI - Make API Call | https://dev-automated-00001.dev.appmixer.ai/designer/ddb55d3e-25d7-428f-a1e8-f7f8593ae669 |

All five are `stage: stopped` — nothing was run.

**Flows** live at `src/appmixer/ai/artifacts/test-flows/` (connector level) and cover all 13 `ai/openai` components:

- **Prompting** — ListModels → SendPrompt → AIAgent
- **Structured Output** — TransformTextToJSON, VariableExtractor, Moderate
- **Embeddings** — GenerateEmbeddings, GenerateEmbeddingsFromFile
- **Media** — GenerateImage → DescribeImages, CreateSpeech → CreateTranscription
- **Make API Call** — MakeApiCall `GET /models` and `POST /chat/completions`

They exercise what this PR changed: SendPrompt's new `instructions` + sampling inputs, AIAgent's `model`/`instructions` as bindable inPorts, and the dynamic outPort options of TransformTextToJSON and VariableExtractor.

**Publish.** `src/appmixer/ai` has no `service.json`, so `appmixer pack ai` is not possible — the `ai` connector is module-rooted. Packed and published `ai/openai` and `ai/gemini` separately. All 13 components changed by this PR were `appmixer remove`d first (12 in openai + `gemini.GenerateEmbeddingsFromFile`) so the instance picked up the new definitions, then re-verified present via `appmixer component ls`.

**Import.** `appmixer e2e import -c ai` (with validation) first failed: `TokenError: Access token not found for component <id>` on the `model` property of every component whose `model` inspector resolves its options through `ListModels`. Cause was `No account ID found — skipping account assignment` (`APPMIXER_SKILL_ACCOUNT_ID` unset), not a missing account — `appmixer:ai:openai` account `69de4000bee6ceff3aeb0f87` already exists. Re-imported with `-a 69de4000bee6ceff3aeb0f87`; all 5 flows now import clean (exit 0, no red-chip variables).

**Also removed** `src/appmixer/ai/openai/artifacts/test-flows/test-flow-e2e-ai-openai-generate-image.json` — it sat inside a module directory, used a `BeforeAll` and carried no `errorHandling`, so it fails the current e2e ruleset. Its GenerateImage coverage is subsumed by the Media flow.

`appmixer e2e validate src/appmixer/ai/artifacts/test-flows` passes all 27 validators. Three advisory warnings remain and are intentional:

- `create-without-cleanup` on the Media flow — false positive; the flow does clean up, via two `appmixer.utils.files.RemoveFile` nodes after AfterAll, but the rule only recognises same-connector Delete/Remove components.
- AIAgent `storeId`/`fileId` untested — they need a pre-existing data store and file.
- VariableExtractor `file` untested — it is mutually exclusive with the `text` input the flow uses.

Determinism: the AIAgent thread id is a per-run `g_uuid4` so conversation history does not accumulate across runs, and every file the flows create (saved text file, generated image, generated speech) is removed at the end.

vtalas (approved):

PR #1243, opened 2026-08-28
ai/groq(major): align with the AI connector contract — rename MakeApiCall, emit `answer`, add lib.js

apx-vero:
Closes Appmixer-ai/appmixer-components#2809

Brings `ai/groq` onto the shared AI connector contract (see Appmixer-ai/appmixer-components#2808) and clears all **13 validator failures** the connector was carrying. `node scripts/validate.js --connector ai` now reports zero groq findings.

### 1. `MakeAPICall` → `MakeApiCall` (breaking)

The misspelling meant the component silently skipped the `makeapicall-standards` rule set, so nothing checked it — and it did need checking. Rebuilt to the standard shape, using `ai/requesty/MakeApiCall` as the reference:

- inputs are now `url` / `method` / `parameters` / `body` / `headers`, with `key-value` inspectors for headers and query parameters and a `textarea` for the body (they were `object`-typed with `type: "text"` inspectors)
- **origin pinning** via a new `lib.resolveApiUrl` modelled on `gladia/lib.js`: the endpoint path is resolved through the WHATWG URL parser and rejected unless it lands on `https://api.groq.com`. That also rejects protocol-relative input like `//example.com/x` and paths carrying credentials, so a mistyped or variable-bound path cannot leak the account's API key to a foreign host (in scope for Appmixer-ai/appmixer-components#2802)
- relative paths stay relative to `/openai/v1/`, so an existing `audio/transcriptions` value still resolves the same way

### 2. `SendPrompt` output shape (breaking)

It emitted the raw API path `choices.[0].message.content` labelled "Response". Every other AI connector emits `answer`, so swapping groq for another provider broke every downstream binding. It now emits `answer` and `prompt` alongside the metadata (`id`, `model`, `created`, `finishReason`, `usage.*`).

Input labels and order are also on-contract now: `Search Prompt` → `Prompt`, `Model ID` → `Model`, and the order is prompt → system instructions → model → conversation ID → Advanced (`instructions` used to sit at index 0, ahead of the prompt). The existing sampling inputs and the conversation-state handling are unchanged.

### 3. `ListModels`

Had no `schema` and no `options` on its outPort, so the variable picker was empty and users fell back to Raw Output. It now uses the standard `outputType` input plus `lib.sendArrayOutput` / `lib.getOutputPortOptions` with a dynamic outPort source, following `ai/requesty/ListModels`. The inspector transform is renamed `modelsToSelectArray` → `toSelectArray` to match the array output (`{ result, count }`), and all three callers were updated.

### 4. `lib.js`

groq was one of three AI connectors without one. Added `getBaseUrl`, `request`, `resolveApiUrl`, `keyValueToObject` and the array-output helpers. Every component and `auth.js` now goes through it instead of re-implementing request handling.

### 5. `CreateTranscription` / `CreateTranslation`

`model` was a `type: "select"` with `whisper-large-v3` and `whisper-large-v3-turbo` baked into the inspector, so the component could not reach a model Groq adds later. Both are now `type: "text"` with a `ListModels` source (typeahead, so a manual value still works if the source fails), keeping `whisper-large-v3` as the default. Both also declare and assert `model` as required.

> The issue only called out `CreateTranscription` for this, but `CreateTranslation` had the identical hardcoded list and the fix is the same — leaving one of the pair broken seemed worse than the small extra diff.

### 6. Icons and output-port examples

Replaced the JPEG data URIs (5 components + `module.json`) with an SVG, and added `example` to the remaining outPort options.

> **Note on the icon:** the previous asset was a rasterised JPEG, so there was nothing to trace. I hand-authored a simple mark in Groq's brand orange (`#F55036`) and checked it into `artifacts/icon.svg` so it is editable. Happy to swap it for the official Groq SVG if one is available.

## Versioning

Breaking (component rename + output field rename) → **major bump to `2.0.0`** in `bundle.json`, `module.json` and `package.json`, with a changelog entry that spells out the migration for existing flows.

## Verification

- `npm run lint` — clean
- `node scripts/validate.js --changed --base origin/dev` — all 25 rules OK
- `node scripts/validate.js --connector ai` — 0 groq findings (was 13); no new findings introduced for the other `ai/*` modules
- `appmixer connector verify` skipped — `src/appmixer/ai/groq/artifacts/samples/` does not exist, and live mode needs credentials that are not available in this job

Publishing and E2E flows are handled in a separate follow-up step; nothing was run against a live instance here.


apx-vero:
Published `appmixer.ai.groq` + imported 3 E2E flows (**NOT started** — all three are `stopped`).

| Flow | Import | Designer link |
|------|--------|---------------|
| E2E AI Groq - API call | ✅ clean | https://dev-automated-00001.dev.appmixer.ai/designer/97148fed-d2e5-4125-bbb8-46200fbd2e9e |
| E2E AI Groq - prompt | ❌ variable check failed | https://dev-automated-00001.dev.appmixer.ai/designer/8441ed67-de45-4918-929b-371ed9d7af3c |
| E2E AI Groq - audio | ❌ variable check failed | https://dev-automated-00001.dev.appmixer.ai/designer/e3ae0835-e82e-42f1-bce3-6f4826364750 |

`appmixer e2e import -c ai` exited **1**. Local `appmixer e2e validate` is clean (27 validators, 0 failures), so this is not a flow-JSON problem — **the instance has no `appmixer:ai:groq` account**:

```
No account ID found — skipping account assignment.
```

`appmixer account ls` confirms it: there are `appmixer:ai:openai`, `:mistral`, `:exa`, `:huggingface`, `:requesty` accounts, but none for `groq`.

With nothing bound, the server-side variables check cannot resolve the `model` typeahead (its inspector `source` calls `ListModels`, which needs auth), so both flows that use a groq component with a dynamic input source fail the check:

```
variable fetch error: 68ce61ef-…-1bcfd1a84dc4 <- d6935ea0-…-63be55daba1.out:
{"static":[{"error":{"message":"Access token not found for component 68ce61ef-…-1bcfd1a84dc4",
"data":{"componentId":"68ce61ef-…-1bcfd1a84dc4"},"code":1,"name":"TokenError"},
"path":{"srcComponentId":"d6935ea0-…-63be55daba1","srcComponentOut":"out",
"tgtComponentId":"68ce61ef-…-1bcfd1a84dc4","tgtComponentIn":"in"},"property":"model"}]}
```

(Same error for `CreateTranslation` in the audio flow and for both `SendPrompt` nodes in the prompt flow. `MakeApiCall` has no dynamic inspector source, which is why the API-call flow passed.)

**Needs a human:** connect a Groq API key as an `appmixer:ai:groq` account in the designer, then re-run `appmixer e2e import -c ai`. I did not create one.

### What was published

`appmixer pack ai` does not work here — `src/appmixer/ai` has no `service.json`, it is a namespace of modules — so the unit was the module: `appmixer pack ai/groq` → `appmixer publish appmixer.ai.groq.zip`. All five components (plus the shared `lib.js`) were `appmixer remove`d first, including the obsolete `appmixer.ai.groq.MakeAPICall`, so the renamed `MakeApiCall` is what is now on the instance.

### Flow coverage

`groq.SendPrompt` ×2 (output contract + conversation memory over a shared Conversation ID), `groq.MakeApiCall` ×2 (relative `GET` path, `POST` with JSON body + key-value headers), `groq.CreateTranscription`, `groq.CreateTranslation`. `groq.ListModels` is `private: true` — a dropdown source helper, not a testable node (`no-private-component-node`) — so `MakeApiCall` covers the same `/models` endpoint instead.

Notes on pinned model IDs, the external audio asset and the three expected validator warnings are in `src/appmixer/ai/artifacts/test-flows/README.md`.

### Unrelated

The import flagged 5 pre-existing `appmixer:ai` flows on the instance as orphans (`E2E AI OpenAI - Embeddings / Media / Prompting / Structured Output / Make API Call`) because no local file claims them. I left them alone — they look like openai-module work that was never committed to `artifacts/test-flows/`.

vtalas:
## Review + release prep (head `8a649d18`)

Reviewed against the AI connector contract (#2808) and the review-connector checklist. The PR does what the issue asks; three things needed fixing before release, all pushed in `8a649d18` on top of a merge of `upstream/dev` (no conflicts).

### Fixed

| Severity | Component | Rule | Finding | Fix |
|---|---|---|---|---|
| error | ListModels | `dynamic-outport-item-schema` | Dynamic outPort but no exported `ITEM_SCHEMA` — the validator on current `dev` fails on it (the PR predates that rule), and `connector verify` would treat every field as required. | Exported `ITEM_SCHEMA` (`required: id/object/created/owned_by`, per-leaf `example`, added `max_completion_tokens`) and pass `ITEM_SCHEMA.properties` to the helper. |
| error | ListModels (+ SendPrompt, CreateTranscription, CreateTranslation sources) | `source.uncached` | `ListModels` backs three model typeaheads, so an inspector open fired one uncached `/models` call per dropdown and any failure popped an error in the designer. | The three sources pass `isSource: true`; on that path `lib.requestCached()` serves the list from `context.staticCache` under `context.lock` (key = url + API key, TTL `context.config.listCacheTTL`, default 120 s) and errors collapse to `{ result: [], count: 0 }` so the user can still type a model ID. Flow-time calls are unchanged (no cache, errors propagate). |
| warning | CreateTranscription, CreateTranslation | `resource.stream-leak` | The file read stream was opened before `getFileInfo` and never destroyed when the upload threw; `temperature: null` crashed on `.toString()`. | Metadata first, stream second; `fileStream.destroy()` on the error path; null-safe temperature. |

Verified with a mock-context harness: `generateOutputPortOptions` never hits the API; second `isSource` call is a cache hit; a different API key is a different cache entry; a 500 on the source path returns an empty list while a 500 at flow time throws; a failed upload destroys the stream; multipart `content-type` from `form-data` wins over the JSON default.

### Checked OK

- `resolveApiUrl`: 22 inputs — relative (`models`, `audio/transcriptions`), host-absolute and full Groq URLs resolve; `https://evil.com`, `//evil.com/x`, `http://api.groq.com` (scheme downgrade), `api.groq.com.evil.com`, `user:pw@`, `api.groq.com@evil.com`, `javascript:` all throw `CancelError`. `..` segments are normalised by the URL parser and stay on the origin.
- Conversation state: same `flow.stateGet/stateSet(conversationId)` key and `{ messages }` shape as 1.0.5, so existing conversations survive the upgrade. The old code logged the full request including the `Authorization` header; the new one does not.
- `bundle.json` changelog ascending (1.0.5 → 2.0.0), `module.json` / `package.json` at 2.0.0, one version on the branch.
- `node scripts/validate.js --connector ai` — 0 groq findings on current `dev` rules; `--changed --base upstream/dev` passes all 28 rules; eslint clean.
- `appmixer e2e validate` on the three flows: 29 validators pass, only the documented warnings (`input-coverage-optional` for `parameters`, `process-config` store IDs injected at upload). Layout follows the 208/128 staircase.

### Not changed (for the record)

- `SendPrompt` keeps `maxCompletionTokens`; the contract names it `maxTokens`. #2809 explicitly said to keep the existing sampling inputs, so left alone — but if we ever want it on-contract, this major bump is the cheap moment.
- The icon is a hand-drawn placeholder, not the official Groq mark.
- `answer` falls back to the literal `[No reply generated]` when the API returns no choice (inherited from 1.0.5).

### Instance state

`appmixer.ai.groq` was removed and republished on `dev-automated-00001` from `8a649d18` (all five components). The three E2E flows are imported and `stopped` / never run.

**Still blocked on the same thing as before: there is no `appmixer:ai:groq` account on the instance and no Groq API key anywhere I can reach.** Once a key is connected in the designer, the remaining steps are `appmixer e2e import -c ai -a <accountId>` → `appmixer e2e run` ×3 → post the report to #2809.

vtalas:
**Unblocked — E2E green on head `ef5bd4f2`.** A Groq key was connected as account `groq e2e test` (`6a9fc153b06549789ce2aee3`) on dev-automated-00001, the module was republished from this commit, all three flows imported cleanly (server-side variable check now resolves the `model` typeaheads through the cached `isSource` path) and passed: API call 6/6 checks, prompt 9/9, audio 4/4. Second `SendPrompt` recalled the token from the shared Conversation ID, so the state handling survived the rewrite.

One more commit was needed: Groq had retired `llama-3.3-70b-versatile`, so both chat flows now pin `openai/gpt-oss-120b` with a 1000-token completion budget (reasoning model), and `ListModels` `ITEM_SCHEMA` declares the extra fields every live record carries. CI is green. Full report with designer links on Appmixer-ai/appmixer-components#2809.

vtalas (approved):

PR #1244, opened 2026-08-28
line(e2e): add the E2E test flow suite (the connector shipped with none)

apx-vero:
## Summary

`line` shipped to `dev` (Appmixer-ai/appmixer-components#2579) with **no `artifacts/` directory and no E2E coverage at all**. This adds the suite.

## Coverage

| flow | components | mode |
|---|---|---|
| `test-flow-insights.json` | `GetNumberOfFollowers`, `GetNumberOfMessageDeliveries` | automatic ✅ |
| `test-flow-makeapicall.json` | `MakeApiCall` (`GET /v2/bot/info`) | automatic ✅ |
| `test-flow-messaging-manual.json` | `NewMessages`, `SendReplyMessage`, `SendPushMessage` | needs one human message to the bot |

Both automatic flows pass against the live account on `dev-automated-00001`.

`test-flow-messaging-manual.json` cannot be self-provoked: LINE only issues a `replyToken` for a real inbound webhook event, and it is single-use with a short TTL. The flow starts, a human sends any text to the bot, `NewMessages` fires, the flow replies to that message and then pushes a second one to the same user. `AfterAll` waits 600s.

## Deliberately not covered

| Component | Reason |
|---|---|
| `SendBroadcastMessage` | Delivers to **every follower of the channel** — 4 real accounts here. An automated test must not message real people; covering it needs a channel with no human followers. |
| `LeaveGroupOrRoom` | Needs the bot inside a group or room, and the call is destructive — the bot cannot re-add itself. Needs a throwaway group and a human re-invite per run. |

`ListRichMenus` and `GetRichMenu` are `private: true` (dynamic-output source helpers, not user-facing actions), so the `no-private-component-node` rule puts them out of scope for E2E — they are exercised through the inspector dropdowns of the components that consume them.

## Known maintenance note

`test-flow-insights.json` pins a literal `date` in `YYYYMMDD`. LINE keeps insight data for 365 days and aggregates it a day in arrears, so the pinned date works for a year and then returns `status: "unready"`. No date-formatting modifier is available to derive `YYYYMMDD` from `g_now`; bumping the literal (or adding a `CodeBlock` that formats it) is the maintenance step. Called out in the README.

## Out of scope here

`node scripts/validate.js --connector line` reports **10 `component-icon-svg` failures** — every component icon is a PNG data URI rather than SVG. Pre-existing on `dev` and not touched here: fixing it needs the official LINE SVG asset, which I don't have, and tracing a brand mark by hand is worse than leaving it. Worth its own issue.

Minor, also left alone: `ListRichMenus` passes `arrayPropertyValue: 'models'` to `lib.sendArrayOutput`, which ignores that argument entirely (it always emits `{result, count}`), and labels its output "Models" — leftovers from whichever connector it was copied from. Harmless at runtime, wrong in the picker label.

### DIFF
--- a/src/appmixer/line/artifacts/test-flows/README.md
+++ b/src/appmixer/line/artifacts/test-flows/README.md
@@ -1,0 +1,52 @@
+# LINE E2E test flows
+
+```bash
+appmixer e2e import src/appmixer/line/artifacts/test-flows --account <accountId>
+appmixer e2e run <flowId> --fix
+```
+
+## Fixtures these flows assume
+
+| Fixture | Value used |
+|---|---|
+| Bot | `@523cawgz` ("Appmixer"), userId `U59f8a8aa5337e9e39bec3461bb71615e` |
+| Insight date | `20260826` — see the note below |
+
+## Fully automatic
+
+| Flow | Covers |
+|---|---|
+| `test-flow-insights.json` | GetNumberOfFollowers, GetNumberOfMessageDeliveries |
+| `test-flow-makeapicall.json` | MakeApiCall (`GET /v2/bot/info`) |
+
+`test-flow-insights.json` pins a literal `date` (`YYYYMMDD`). LINE keeps insight
+data for 365 days and only aggregates it a day in arrears, so the pinned date
+works for a year and then starts returning `status: "unready"`. There is no
+date-formatting modifier available to compute `YYYYMMDD` from `g_now`, so bumping
+the literal (or adding a CodeBlock that formats it) is the maintenance step.
+
+## Needs one manual action
+
+### `test-flow-messaging-manual.json`
+
+Covers **NewMessages → SendReplyMessage → SendPushMessage**. LINE only issues a
+`replyToken` for a real inbound webhook event, and the token is single-use with a
+short TTL, so the chain cannot be provoked from inside the flow.
+
+1. Start the flow.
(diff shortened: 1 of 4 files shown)


apx-vero:
E2E report: Appmixer-ai/appmixer-components#2579 (comment https://github.com/Appmixer-ai/appmixer-components/issues/2579#issuecomment-5454535956).

Both automatic flows pass on `dev-automated-00001`; the manual messaging flow is imported and waiting at [809fa947](https://dev-automated-00001.dev.appmixer.ai/designer/809fa947-d04f-4ed9-be1d-e04dbde6c64b) for someone to message the bot.

apx-vero:
## Icons fixed

The 10 `component-icon-svg` failures called out as out of scope are now fixed — the user supplied the official LINE SVG, so all 10 component icons plus `service.json` carry it as an `svg+xml` data URI (`32ad8398`). Only the icon value changed in each file; formatting is untouched, so the diff is 11 lines. Bundle bumped to `2.0.1`.

`node scripts/validate.js --connector line`: **11 failures → 1**. The remaining one is the `ListRichMenus` `ignoreAuth` advisory, deliberately left — that source calls the LINE API and needs `context.auth`, so the flag would break it.

Republished and re-ran both automatic flows after the swap: **still 2/2 PASSED**.

vtalas (approved):

